Issue #20, opened 2026-04-12
Admin plan management endpoint (beta manual upgrades)


dr14-make:
Part of #11. Depends on #TBD-db.

## Goal

A tiny admin API for manually setting a user's plan tier during the beta, before Stripe integration exists. Not user-facing.

### Route

\`\`\`
POST /admin/plans/:user_id
Authorization: Bearer <RELAY_ADMIN_SECRET>
Content-Type: application/json

{ "tier": "pro" }
\`\`\`

- Auth: constant-time comparison against `RELAY_ADMIN_SECRET` env var. Fail boot if the env var is missing **and** `NODE_ENV=production`.
- Writes via `db.setPlan(user_id, tier)` which sets the row to the documented defaults for that tier.
- Idempotent. Returns the resulting plan row as JSON.

### Convenience

- `GET /admin/plans/:user_id` — returns current plan, same auth
- `GET /admin/users?github_login=dr14` — lookup user by GH login so the operator doesn't need the numeric id, same auth

## Why not a CLI tool

A one-endpoint HTTP API is easier to curl from anywhere, easier to audit (access logs), and doesn't require shell access to the production host. If operators want a CLI wrapper, that's a 5-line bash alias.

## Acceptance

- [ ] All admin routes return 401 without the bearer token
- [ ] All admin routes return 401 with a wrong token (timing-safe compare)
- [ ] Setting a valid tier updates the row and returns it
- [ ] Setting an invalid tier returns 400 with the list of valid tiers
- [ ] Route registered under a clearly-named subrouter that's easy to remove when Stripe replaces it
- [ ] Boot fails in production if `RELAY_ADMIN_SECRET` is unset

## Parallelizable with

#5, #6, #7. ~2-3 hours of work.

Issue #18, opened 2026-04-12
Broker: gate OAuth provider access by plan tier


dr14-make:
Part of #11. Depends on #TBD-db.

## Goal

When a daemon initiates an OAuth flow through the broker (`GET /auth/:provider`), reject providers that aren't included in the daemon owner's plan.

## Current state

[src/broker/router.ts:106-119](../blob/main/src/broker/router.ts#L106-L119) — already verifies ECDSA sig, timestamp, that the daemon UUID is currently connected. Missing: any check that the *user* is allowed to use this provider.

## Logic

After the existing daemon-presence check, add:

\`\`\`ts
if (process.env.RELAY_REQUIRE_CLAIM !== "false") {
  const daemon = db.getDaemon(relay_uuid);
  if (!daemon) return res.status(403).json({ error: "daemon-not-claimed" });
  const plan = db.getPlan(daemon.user_id);
  const allowed = plan.oauth_providers; // '*' | string[]
  if (allowed !== "*" && !allowed.includes(provider)) {
    return res.status(403).json({
      error:    "provider-not-in-plan",
      provider,
      tier:     plan.tier,
      upgrade:  \`\${BASE_URL}/dashboard/billing\`,
    });
  }
}
\`\`\`

## Plan defaults

From the epic:
- `free`: `['github']`
- `pro`:  `'*'`
- `team`: `'*'`

Stored in `plans.oauth_providers` as JSON-encoded array, or the literal string `*`. Document both forms in the schema comment.

## Acceptance

- [ ] Free-tier user requesting `/auth/slack` → 403 with `provider-not-in-plan`, upgrade URL
- [ ] Free-tier user requesting `/auth/github` → passes through to grant
- [ ] Pro-tier user requesting any provider → passes through
- [ ] Unknown provider (not in PROVIDER_CONFIGS) still returns the existing 400, not the new 403
- [ ] Test matrix in `tests/broker/auth.test.ts`

## Parallelizable with

#5, #6, #9.

Issue #17, opened 2026-04-12
Webhook forwarding: per-user monthly hook quota (429 when exceeded)


dr14-make:
Part of #11. Depends on #TBD-db.

## Goal

Enforce `hook_quota_monthly` from the user's plan on every inbound webhook that gets forwarded to a daemon.

## Where

The Express handler that forwards `/u/:uuid/hooks/*` to the daemon over the WS tunnel — currently around [src/index.ts](../blob/main/src/index.ts) (wired to `relayServer.forward(...)`). Add the check **before** calling `forward()`.

## Logic

\`\`\`ts
if (process.env.RELAY_REQUIRE_CLAIM !== "false") {
  const daemon = db.getDaemon(uuid);
  if (!daemon) return res.status(404).json({ error: "unknown-daemon" });
  const plan  = db.getPlan(daemon.user_id);
  const count = db.incrementHookUsage(daemon.user_id, currentPeriod());
  if (count > plan.hook_quota_monthly) {
    return res.status(429).json({
      error:   "hook-quota-exceeded",
      limit:   plan.hook_quota_monthly,
      period:  currentPeriod(),
      tier:    plan.tier,
    });
  }
}
await relayServer.forward(uuid, ...);
\`\`\`

## `incrementHookUsage` semantics

Single SQL statement, atomic, lazy period reset:

\`\`\`sql
INSERT INTO hook_usage (user_id, period, count) VALUES (?, ?, 1)
ON CONFLICT(user_id, period) DO UPDATE SET count = count + 1
RETURNING count;
\`\`\`

No cron needed — a new row appears when the first hook of a new month arrives. Old rows can be purged later with a background sweep (out of scope).

## Counting semantics

**Increment before forwarding**, so a successful 429 doesn't cost a count but a genuine forward does even if the daemon errors. Rationale: we're billing for *attempted* forwards, which is what consumes relay capacity. Document this in a comment.

Edge case: if `forward()` throws `ClientNotConnectedError`, should we decrement? No — the user's daemon being offline is their problem, not the relay's cost to bear. Count it. Document this too.

## `currentPeriod()` helper

\`\`\`ts
function currentPeriod(now = new Date()): string {
  return \`\${now.getUTCFullYear()}-\${String(now.getUTCMonth() + 1).padStart(2, "0")}\`;
}
\`\`\`

UTC, not server local time, so behaviour is deterministic across regions.

## Acceptance

[...]

Issue #16, opened 2026-04-12
Handshake enforcement: reject unclaimed daemons + concurrent daemon limit


dr14-make:
Part of #11. Depends on #TBD-db.

## Goal

Reject relay handshakes from daemons that aren't claimed by a user, and enforce the `concurrent_daemons` limit from the user's plan.

## Where

[src/relay/server.ts:268-273](../blob/main/src/relay/server.ts#L268-L273) — inside `handleHello`, **after** ECDSA signature verification succeeds (step 5 in the CLAUDE.md handshake spec) but **before** the client is added to the registry.

## Logic

\`\`\`ts
// existing: sig verified, uuid is trustworthy

if (process.env.RELAY_REQUIRE_CLAIM !== "false") {
  const daemon = db.getDaemon(uuid);
  if (!daemon) {
    sendError(ws, "daemon-not-claimed", {
      hint: "Run `dicode relay login` on this machine",
    });
    return ws.close();
  }
  const plan = db.getPlan(daemon.user_id);
  const currentUuids = registry.uuidsForUser(daemon.user_id);
  if (currentUuids.size >= plan.concurrent_daemons) {
    sendError(ws, "concurrent-daemon-limit", {
      limit: plan.concurrent_daemons,
      tier:  plan.tier,
    });
    return ws.close();
  }
  db.touchDaemon(uuid); // update last_seen
}

// existing: add to registry, send welcome
\`\`\`

## Registry changes

Add an index to the in-memory registry so `uuidsForUser(userId)` is O(1):

\`\`\`ts
private byUser = new Map<number, Set<string>>();
\`\`\`

Update on add/remove. Tests must cover: add, remove, multiple daemons per user, user with zero daemons.

## Rollout safety flag

`RELAY_REQUIRE_CLAIM` defaults to **true**. Operators flip it to `false` during initial rollout to avoid breaking existing unclaimed daemons while users migrate. Log a warning every hour if it's `false` in production so it can't quietly stay off.

## Error messages (extend protocol)

Extend the `error` message schema in `src/relay/protocol.ts` with an optional `code` field (`daemon-not-claimed`, `concurrent-daemon-limit`, ...) so the Go client can distinguish these from crypto failures and show actionable help.

## Acceptance

- [ ] Unclaimed daemon → `error { code: "daemon-not-claimed" }`, connection closed
- [ ] Over-limit daemon → `error { code: "concurrent-daemon-limit" }`, connection closed
[...]

PR #22, opened 2026-04-12
feat(db): SQLite persistence layer for users, daemons, plans, hook usage

### LINKED ISSUE #12: SQLite persistence layer + schema + migrations
Part of #11. **Foundation — blocks most other issues.**

## Goal

Introduce a persistent data layer to dicode-relay for users, daemons, plans, and hook usage counters.

## Why `better-sqlite3`

- Sync API — no async round-trip on the WS handshake hot path.
- Single file, trivial Docker volume mount, zero network failure modes.
- Matches project rule "no runtime deps beyond `node:crypto`" better than Supabase/Postgres.
- Drop-in migration to Postgres if ever needed (1–2 days).

## Scope

- Add dep: `better-sqlite3` + `@types/better-sqlite3`.
- Env var: `DICODE_RELAY_DB` (default `/var/lib/dicode/relay.db`, fall back to `./data/relay.db` in dev).
- New module `src/db/index.ts` — opens the DB with `PRAGMA journal_mode = WAL; PRAGMA foreign_keys = ON;`, runs migrations, exposes a small typed API (not a raw handle).
- Migrations folder `src/db/migrations/` with `001_init.sql`. Track applied migrations in a `schema_version` table, run pending on startup.

## Schema

\`\`\`sql
CREATE TABLE users (
  id                           INTEGER PRIMARY KEY AUTOINCREMENT,
  github_id                    INTEGER UNIQUE NOT NULL,
  github_login                 TEXT NOT NULL,
  email                        TEXT,
  github_access_token_encrypted BLOB,   -- nullable; set when user consents to git reuse
  created_at                   INTEGER NOT NULL
);

CREATE TABLE daemons (
  uuid        TEXT PRIMARY KEY,          -- 64 lowercase hex, pubkey-derived
  user_id     INTEGER NOT NULL REFERENCES users(id) ON DELETE CASCADE,
  label       TEXT,
  first_seen  INTEGER NOT NULL,
  last_seen   INTEGER NOT NULL
);
CREATE INDEX daemons_user_id ON daemons(user_id);

CREATE TABLE plans (
  user_id              INTEGER PRIMARY KEY REFERENCES users(id) ON DELETE CASCADE,
  tier                 TEXT NOT NULL CHECK (tier IN ('free','pro','team')),
  hook_quota_monthly   INTEGER NOT NULL,
  concurrent_daemons   INTEGER NOT NULL,
  oauth_providers      TEXT NOT NULL,     -- JSON array or '*' sentinel
  stripe_sub_id        TEXT,
  renews_at            INTEGER
);

CREATE TABLE hook_usage (
  user_id INTEGER NOT NULL REFERENCES users(id) ON DELETE CASCADE,
  period  TEXT NOT NULL,                  -- 'YYYY-MM'
  count   INTEGER NOT NULL DEFAULT 0,
  PRIMARY KEY (user_id, period)
);
\`\`\`

## Typed API surface (what callers see)

\`\`\`ts
export interface Db {
  // users
  upsertUserFromGithub(gh: GithubProfile, token?: Buffer): User;
  getUserById(id: number): User | null;

  // daemons
  getDaemon(uuid: string): DaemonRow | null;
  claimDaemon(uuid: string, userId: number, label?: string): void;
  countActiveDaemonsForUser(userId: number, uuids: Iterable<string>): number;

  // plans
  getPlan(userId: number): Plan;   // returns 'free' defaults if missing
  setPlan(userId: number, tier: Tier): void;

  // usage
  incrementHookUsage(userId: number, period: string): number; // returns new count
}
\`\`\`

`countActiveDaemonsForUser` takes the set of currently-connected UUIDs from the in-memory registry — the DB doesn't know who is connected right now.

## Acceptance

- [ ] `npm run test` passes with new `tests/db/*.test.ts` (round-trip each API method, foreign-key cascade, migration idempotency)
- [ ] DB file is created on first boot if missing
- [ ] `PRAGMA foreign_keys=ON` verified in an init test
- [ ] Dockerfile documents the volume mount path
- [ ] No raw SQL leaks out of `src/db/` — callers use the typed API only

## Out of scope

Encryption of `github_access_token_encrypted` — ship the column nullable; the actual encryption helper ships with #2.

## Parallelizable with

Nothing — this blocks #2, #5, #6, #7, #9. Start here.

dr14-make:
## Summary

Introduces a typed SQLite persistence layer for dicode-relay, backed by `better-sqlite3`. This is the foundation of epic #11 — it unblocks Wave 2 work (#13, #14, #16, #17, #18, #20).

- `src/db/index.ts` — `openDb()` + `Db` interface with cached prepared statements; callers never see raw SQL.
- `src/db/migrations/001_init.sql` — `users`, `daemons`, `plans`, `hook_usage` tables exactly as specified in #12.
- `schema_version` table + idempotent numbered migration runner (`NNN_name.sql`).
- WAL + foreign keys pragmas verified at open time.
- `scripts/copy-assets.mjs` — copies migration SQL into `dist/` during build (tsc does not copy non-TS assets).
- `Dockerfile` — declares `/var/lib/dicode` volume, sets `DICODE_RELAY_DB=/var/lib/dicode/relay.db`, chowns to the `dicode` user, installs native-build toolchain to compile `better-sqlite3`.
- `tests/db/db.test.ts` — 23 tests, real SQLite in `os.tmpdir()`, zero mocks.

## Acceptance criteria (issue #12)

- [x] `npm run test` passes with new `tests/db/*.test.ts` (round-trips every API method, foreign-key cascade, migration idempotency)
- [x] DB file is created on first boot if missing (`openDb` ensures parent dir + creates file)
- [x] `PRAGMA foreign_keys = ON` verified in an init test (and asserted on open)
- [x] Dockerfile documents the volume mount path (`VOLUME ["/var/lib/dicode"]`)
- [x] No raw SQL leaks out of `src/db/` — `Db` is the only export surface

## Design decisions (not explicitly in the issue — call-outs for Wave 2)

- `Plan.oauthProviders` is typed as `string[] | "*"` with `"*"` stored as a sentinel string in SQL. Wave 2 broker-gate (#18) should branch on `=== "*"` for the unrestricted case.
- `getPlan` returns `free` defaults **without inserting a row**, so the `plans` table only contains explicitly set non-free rows. `setPlan` always resets quota/daemon limits/providers to the tier's canonical defaults (no partial edits — ship admin overrides as a later migration).
- `upsertUserFromGithub` preserves the existing encrypted token when the caller omits one, so `#13` can safely update login/email on every OAuth hit without clobbering the stored token.
- `claimDaemon` uses `ON CONFLICT(uuid) DO UPDATE` — re-claiming refreshes `last_seen` and lets a user transfer a UUID between labels; if you want stricter "one owner forever" semantics, add a guard in the caller.
- `countActiveDaemonsForUser` takes the active UUID set from the in-memory relay registry (caller passes it in) — the DB does not know connectivity state.
- Migration runner lives entirely in-process at `openDb` time — no separate CLI step. Add new migrations as `src/db/migrations/NNN_name.sql` and they'll apply on next boot.
- Tier quota/daemon/provider defaults are hard-coded in `PLAN_DEFAULTS` in `src/db/index.ts`. If Wave 2 needs admin-tunable values per user, add columns to `plans` and loosen `setPlan`.

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm run lint` — clean
- [x] `npm run format:check` — clean
- [x] `npm run test:coverage` — 101/101 tests passing, `src/db/index.ts` at **95.02% lines / 100% functions** (threshold 90/90)
- [x] `npm run build` — produces `dist/src/db/migrations/001_init.sql`

## Local CI evidence

```
> npm run typecheck && npm run lint && npm run format:check && npm run test:coverage && npm run build

> tsc --noEmit          # clean
> eslint src tests      # clean
> prettier --check      # All matched files use Prettier code style!

 Test Files  11 passed (11)
      Tests  101 passed (101)

 % Coverage report from v8
---------------|---------|----------|---------|---------|-----------------------
File           | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
---------------|---------|----------|---------|---------|-----------------------
All files      |   93.56 |     82.1 |     100 |   93.56 |
 db            |   95.02 |    87.93 |     100 |   95.02 |
  index.ts     |   95.02 |    87.93 |     100 |   95.02 | ...22,387-388,424-425
---------------|---------|----------|---------|---------|-----------------------

> tsc --project tsconfig.build.json && node scripts/copy-assets.mjs
copy-assets: copied src/db/migrations -> dist/src/db/migrations
```

Closes #12

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,34 @@
 FROM node:25-alpine AS build
 WORKDIR /app
+# build-base + python3 are needed to compile the better-sqlite3 native addon.
+RUN apk add --no-cache build-base python3
 COPY package*.json ./
 RUN npm ci
 COPY tsconfig*.json ./
+COPY scripts ./scripts
 COPY src ./src
 RUN npm run build
 
 FROM node:25-alpine AS runtime
-RUN addgroup -S dicode && adduser -S dicode -G dicode
+# build tools needed once to compile better-sqlite3 during `npm ci --omit=dev`;
+# pruned from the final image layer at the end.
+RUN apk add --no-cache --virtual .build-deps build-base python3 \
+    && addgroup -S dicode && adduser -S dicode -G dicode
 WORKDIR /app
 COPY package*.json ./
-RUN npm ci --omit=dev
+RUN npm ci --omit=dev && apk del .build-deps
 COPY --from=build /app/dist ./dist
+
+# Persistence layer: SQLite database lives under /var/lib/dicode. Mount a
+# Docker volume here so the DB survives container restarts. The default
+# DICODE_RELAY_DB path matches this directory.
+#
+#   docker run -v dicode-relay-data:/var/lib/dicode ...
+#
+RUN mkdir -p /var/lib/dicode && chown -R dicode:dicode /var/lib/dicode
+VOLUME ["/var/lib/dicode"]
+ENV DICODE_RELAY_DB=/var/lib/dicode/relay.db
+
 USER dicode
 EXPOSE 5553
 CMD ["node", "dist/index.js"]
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,7 @@
       "name": "dicode-relay",
       "version": "0.1.0",
       "dependencies": {
+        "better-sqlite3": "^12.8.0",
         "express": "^5.0.1",
         "grant": "^5.4.22",
         "uuid": "^11.0.3",
@@ -16,6 +17,7 @@
       },
       "devDependencies": {
         "@eslint/js": "^9.17.0",
+        "@types/better-sqlite3": "^7.6.13",
         "@types/express": "^5.0.0",
         "@types/node": "^22.10.0",
         "@types/uuid": "^10.0.0",
@@ -1170,6 +1172,16 @@
         "win32"
       ]
     },
+    "node_modules/@types/better-sqlite3": {
+      "version": "7.6.13",
+      "resolved": "https://registry.npmjs.org/@types/better-sqlite3/-/better-sqlite3-7.6.13.tgz",
+      "integrity": "sha512-NMv9ASNARoKksWtsq/SHakpYAYnhBrQgGD8zkLYk/jaK8jUGn08CfEdTRgYhMypUQAfzSP8W6gNLe0q19/t4VA==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@types/node": "*"
+      }
+    },
     "node_modules/@types/body-parser": {
       "version": "1.19.6",
       "resolved": "https://registry.npmjs.org/@types/body-parser/-/body-parser-1.19.6.tgz",
@@ -1846,6 +1858,60 @@
       "dev": true,
       "license": "MIT"
     },
+    "node_modules/base64-js": {
+      "version": "1.5.1",
+      "resolved": "https://registry.npmjs.org/base64-js/-/base64-js-1.5.1.tgz",
+      "integrity": "sha512-AKpaYlHn8t4SVbOHCy+b5+KKgvR4vrsD8vbvrbiQJps7fKDTkjkDry6ji0rUJjC0kzbNePLwzxq8iypo41qeWA==",
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/feross"
+        },
+        {
+          "type": "patreon",
+          "url": "https://www.patreon.com/feross"
+        },
+        {
+          "type": "consulting",
+          "url": "https://feross.org/support"
+        }
+      ],
+      "license": "MIT"
+    },
+    "node_modules/better-sqlite3": {
+      "version": "12.8.0",
+      "resolved": "https://registry.npmjs.org/better-sqlite3/-/better-sqlite3-12.8.0.tgz",
+      "integrity": "sha512-RxD2Vd96sQDjQr20kdP+F+dK/1OUNiVOl200vKBZY8u0vTwysfolF6Hq+3ZK2+h8My9YvZhHsF+RSGZW2VYrPQ==",
+      "hasInstallScript": true,
+      "license": "MIT",
+      "dependencies": {
+        "bindings": "^1.5.0",
+        "prebuild-install": "^7.1.1"
+      },
+      "engines": {
+        "node": "20.x || 22.x || 23.x || 24.x || 25.x"
+      }
+    },
+    "node_modules/bindings": {
+      "version": "1.5.0",
+      "resolved": "https://registry.npmjs.org/bindings/-/bindings-1.5.0.tgz",
+      "integrity": "sha512-p2q/t/mhvuOj/UeLlV6566GD/guowlr0hHxClI0W9m7MWYkL1F0hLo+0Aexs9HSPCtR1SXQ0TD3MMKrXZajbiQ==",
+      "license": "MIT",
+      "dependencies": {
+        "file-uri-to-path": "1.0.0"
+      }
+    },
+    "node_modules/bl": {
+      "version": "4.1.0",
+      "resolved": "https://registry.npmjs.org/bl/-/bl-4.1.0.tgz",
+      "integrity": "sha512-1W07cM9gS6DcLperZfFSj+bWLtaPGSOHWhPiGzXmvVJbRLdG82sH/Kn8EtW1VqWVA54AKf2h5k5BbnIbwF3h6w==",
+      "license": "MIT",
+      "dependencies": {
+        "buffer": "^5.5.0",
+        "inherits": "^2.0.4",
+        "readable-stream": "^3.4.0"
+      }
+    },
     "node_modules/bn.js": {
       "version": "4.12.3",
       "resolved": "https://registry.npmjs.org/bn.js/-/bn.js-4.12.3.tgz",
@@ -1894,6 +1960,30 @@
       "integrity": "sha512-cKV8tMCEpQs4hK/ik71d6LrPOnpkpGBR0wzxqr68g2m/LB2GxVYQroAjMJZRVM1Y4BCjCKc3vAamxSzOY2RP+w==",
       "license": "MIT",
       "optional": true
+    },
+    "node_modules/buffer": {
+      "version": "5.7.1",
+      "resolved": "https://registry.npmjs.org/buffer/-/buffer-5.7.1.tgz",
+      "integrity": "sha512-EHcyIPBQ4BSGlvjB16k5KgAJ27CIsHY/2JBmCRReo48y9rQ3MaUzWX3KVlBa4U7MyX02HdVj0K7C3WaB3ju7FQ==",
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/feross"
+        },
+        {
+          "type": "patreon",
+          "url": "https://www.patreon.com/feross"
+        },
+        {
+          "type": "consulting",
+          "url": "https://feross.org/support"
+        }
+      ],
+      "license": "MIT",
+      "dependencies": {
+        "base64-js": "^1.3.1",
+        "ieee754": "^1.1.13"
+      }
     },
     "node_modules/buffer-equal-constant-time": {
       "version": "1.0.1",
@@ -2004,6 +2094,12 @@
         "node": ">= 16"
       }
     },
+    "node_modules/chownr": {
+      "version": "1.1.4",
+      "resolved": "https://registry.npmjs.org/chownr/-/chownr-1.1.4.tgz",
+      "integrity": "sha512-jJ0bqzaylmJtVnNgzTeSOs8DPavpbYgEr/b0YL8/2GO3xJEhInFmhKMUnEJQjZumK7KXGFhUy89PrsJWlakBVg==",
+      "license": "ISC"
+    },
     "node_modules/color-convert": {
       "version": "2.0.1",
       "resolved": "https://registry.npmjs.org/color-convert/-/color-convert-2.0.1.tgz",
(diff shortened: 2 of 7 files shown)


dr14-make (commented):
## Summary
Solid foundation — schema matches #12 exactly, API surface covers everything Wave 2 needs, real-SQLite tests with FK cascade + migration idempotency + atomic increment, CI green, coverage 95.02% on `src/db/`. A few non-blocking call-outs around upsert semantics for Wave 2 consumers.

## Blocking
(none)

## Suggestions (non-blocking)

- **src/db/index.ts:1013-1019** — `upsertUserFromGithub` unconditionally sets `email = excluded.email`. When Wave 2 (#13) calls this on every OAuth login, a caller that omits `email` (becomes `null` via `gh.email ?? null`) will clobber a previously-stored email. The token column uses `COALESCE` precisely to avoid this footgun — consider the same for `email` (`COALESCE(excluded.email, users.email)`), or document that callers must always pass the current email. The PR body says "preserves the existing encrypted token when the caller omits one" but doesn't mention email gets overwritten — worth aligning behavior or docs.

- **src/db/index.ts:1017-1019** — the `COALESCE` preservation of `github_access_token_encrypted` means there is no way to ever clear a token (e.g. on revoke) through `upsertUserFromGithub`. Probably fine for Wave 2's use case (set-only), but if #13 ever needs a clear path, a dedicated `clearUserToken(id)` method would be cleaner than overloading upsert.

- **src/db/index.ts:1121-1131** — `incrementHookUsage` does an UPSERT then a SELECT inside a `db.transaction`. Because better-sqlite3 is synchronous, the transaction is unnecessary for atomicity — a single `INSERT ... ON CONFLICT ... DO UPDATE ... RETURNING count` would be one round trip and remove the "row disappeared" defensive error. Not a correctness issue, purely a simplification.

- **src/db/index.ts:1108-1119** — `setPlan` hard-resets `stripe_sub_id` and `renews_at` to `NULL` on every call via the UPSERT. This matches the PR body's "no partial edits" note, but it means calling `setPlan(u, 'pro')` after a Stripe webhook has set the subscription ID will nuke the stripe linkage. Wave 2 (#18/#20) will need to set these, so expect to revisit — probably fine to ship, just flagging.

- **src/db/index.ts:1082-1091** — `countActiveDaemonsForUser` fetches all of a user's daemon UUIDs and filters client-side. For a user with thousands of registered daemons this scales linearly with owned rather than active; for current scale it's fine, but a `WHERE uuid IN (?, ?, ...)` with a dynamically bound list would be O(active). Nit.

- **tests/db/db.test.ts:1483-1496** — the "atomic under concurrent Promise.all" test doesn't actually exercise concurrency in a meaningful way: better-sqlite3 is fully synchronous so all 200 `incrementHookUsage` calls execute serially on the same tick regardless of `Promise.all`. The test still proves the UPSERT semantics are correct (1..N unique counts) but the name oversells it. Consider renaming to "returns a unique sequential count per call" or similar.

## Nits

- **src/db/index.ts:1161-1164** — `fk[0]?.foreign_keys !== 1` could use `handle.pragma('foreign_keys', { simple: true })` for a cleaner single-value read, but the current approach is fine.

- **scripts/copy-assets.mjs:723-724** — the output path `dist/src/db/migrations` includes the `src/` prefix. That's because `tsconfig.build.json` preserves `rootDir` layout only when multiple roots exist; worth double-checking that `dist/index.js` resolves `migrations/` correctly at runtime via `fileURLToPath(import.meta.url)`. The PR body confirms the build output location so presumably verified, just flagging.

- **src/db/index.ts:906** — migration file regex `/^(\d{3})_[a-z0-9_]+\.sql$/` — fine, just note that migrations jumping versions (e.g. 001, 003) would apply silently without error. If you want strict sequentiality, add an assertion.

- **tests/db/db.test.ts:1236** — `db: null as unknown as Db` works but `db: undefined as unknown as Db` or a lazy init pattern is slightly cleaner. Very minor.

## What looks great

- Prepared statements cached on instance construction, never re-prepared per call.
- No `any`, no `@ts-ignore`, no raw DB handle exported.
- `PLAN_DEFAULTS` typed with `Record<Tier, PlanDefaults>`.
- `parseOauthProviders` validates shape before returning — won't silently return a broken array.
- Migration runner wraps each migration in a transaction so partial application can't happen.
- Dockerfile prunes build toolchain in the same layer as `npm ci --omit=dev` — image stays small.
- Coverage 95.02% / 100% functions, well above the 90% threshold.
- All Wave 2 API methods (`upsertUserFromGithub`, `getDaemon`, `claimDaemon`, `countActiveDaemonsForUser`, `getPlan`, `setPlan`, `incrementHookUsage`) are present with the signatures issue #12 specified.

dr14-make (commented):
## Security review

Scope: threats against a malicious daemon (valid unclaimed keypair) or a malicious external HTTP caller. Env-var / filesystem-level threats against a trusted operator are out of scope per the task.

### Critical (blocking)

None.

### High

None in this PR, but one Wave 2 prerequisite that is load-bearing and worth calling out so it is not lost:

- `src/db/index.ts:1024-1031` — `claimDaemon` uses `ON CONFLICT(uuid) DO UPDATE SET user_id = excluded.user_id`. This means at the DB layer a call with a UUID already owned by another user silently transfers ownership. The PR description correctly notes this is a deliberate caller-side responsibility, but I want it on the record: **Wave 2 (#16 / handshake path) MUST gate `claimDaemon` so that a second user cannot re-bind a UUID already claimed by a different `user_id`**, otherwise anyone who generates a keypair matching an already-claimed UUID (or more realistically, a compromised daemon key) can silently hijack quota/plan attribution. Consider adding a `claimDaemonStrict` helper now that errors on cross-user rebinding, so the caller does not have to remember.

### Low / advisory

- `src/db/index.ts:1151-1168` — `openDb` does not `fs.chmodSync(path, 0o600)` after creation. better-sqlite3 creates the file at the process umask, typically 0644 → world-readable. In the shipped Docker image only the `dicode` user exists so practical impact is nil today, but once the `github_access_token_encrypted` column is populated (#13) and if the image is ever run on a host with other users, or the DB path is bind-mounted, the file becomes a plaintext-email / ciphertext leak surface. One-line fix: `fs.chmodSync(path, 0o600)` right after `new Database(path)`. Also worth restricting the parent dir to `0o700` on `mkdirSync`.

- `src/db/index.ts:1082-1091` — `countActiveDaemonsForUser` loads *every* daemon row for the user into memory before intersecting with the active set. Plan enforcement caps `concurrent_daemons` at 10 (team tier), so in the happy path this is fine, but at the DB layer itself there is no `LIMIT` and no cap on rows in `daemons` for a given `user_id`. A user who has historically claimed thousands of UUIDs (which the current `claimDaemon` happily allows with no upper bound — every new handshake from a new keypair inserts a row) could make this query grow unboundedly. Not exploitable today, but worth either (a) flipping the query to `SELECT COUNT(*) FROM daemons WHERE user_id = ? AND uuid IN (...)` with a parameterised IN-list, or (b) adding a periodic GC that deletes daemons with `last_seen < now - 90d`. Advisory only — does not block.

- `src/db/index.ts:1034-1042` — `setPlan`'s `ON CONFLICT DO UPDATE` does **not** touch `stripe_sub_id` or `renews_at`, so downgrading pro→free leaves the old stripe subscription id stranded in the row. Not a security issue, but may confuse the billing reconciliation in #17. Worth a comment or an explicit reset.

### Not found (positive confirmations)

- **SQL injection**: every query uses `?` placeholders bound via `better-sqlite3` prepared statements. No string concatenation of user input anywhere. The only `db.exec` call with non-literal SQL is in `runMigrations`, and that SQL comes from files on disk inside the image, not from any request path. ✓
- **Migration filename injection**: `MIGRATION_FILE_RE = /^(\d{3})_[a-z0-9_]+\.sql$/` strictly anchors the name, and the `version` parameter to the `INSERT INTO schema_version` statement is parameter-bound, not interpolated. An attacker-controlled file in the migrations dir still cannot inject SQL via the filename itself. ✓
- **FK cascade actually enforced**: `PRAGMA foreign_keys = ON` at open, re-asserted with a `pragma("foreign_keys")` read-back that throws on mismatch (`src/db/index.ts:1161-1164`). The `foreign-key cascade` test (`tests/db/db.test.ts:1499-1521`) deletes a user on a fresh raw handle and verifies daemons/plans/hook_usage all drop to 0. No orphaned-daemon bypass. ✓
- **Default plan does not write**: `getPlan` returns the free-tier struct without running any INSERT when the row is absent (`src/db/index.ts:1093-1106`), and the test at `tests/db/db.test.ts:1426-1441` explicitly asserts the `plans` table still has 0 rows after the lookup. A malicious daemon cannot force a `plans` row to exist just by triggering a plan read. ✓
- **Atomic usage counter**: `incrementHookUsage` uses `INSERT ... ON CONFLICT(user_id, period) DO UPDATE SET count = hook_usage.count + 1`, which is atomic at the SQLite level; the surrounding `handle.transaction()` serialises the run+get under a single implicit BEGIN/COMMIT, and better-sqlite3 is synchronous per-process so JS microtask interleaving between two `Promise.all` calls is impossible. The concurrent-hits test (`tests/db/db.test.ts:1483-1496`) asserts 200 increments produce exactly the unique set {1..200}. No TOCTOU. ✓
- **Safe PRAGMAs**: only `journal_mode = WAL` and `foreign_keys = ON` are set. `synchronous` stays at the WAL default (`NORMAL`), not `OFF`, so committed rows survive a crash. ✓
- **No secret logging**: `src/db/index.ts` contains zero logger/console calls. Row objects (including `github_access_token_encrypted`) are never passed to any logger in the PR. ✓
- **Unbounded queries**: no `SELECT * FROM hook_usage` / `SELECT * FROM daemons` without a WHERE clause. Every read is keyed by `user_id`, `uuid`, or `(user_id, period)`. ✓
- **Parent-dir traversal**: `mkdirSync(dirname(path), { recursive: true })` is only a concern if `DICODE_RELAY_DB` is attacker-controlled, which is out of scope for the declared threat model (malicious daemon / external HTTP caller). ✓

Verdict: **no blocking findings**. The one High item is a documentation-of-intent nudge for Wave 2, not a bug in this PR. Two Low advisories (chmod 0600, bounded daemon query) would be nice to land before #13 adds the token column but are not required to merge.

dr14-make:
Follow-up commit `1c002c6` addresses the review findings:

- **Security (low, load-bearing)** — DB file chmod'd to 0600 on create; also self-heals loose perms on open. Two tests (skipped on win32).
- **Security (high)** — `claimDaemonStrict` + exported `DaemonAlreadyClaimedError`. #14 should use the strict variant so rebinding a UUID to a different user surfaces as 409. Existing `claimDaemon` retained unchanged so callers that want upsert semantics still have them. Six new tests covering all four branches.
- **Code review #1 — email clobber** — `upsertUserFromGithub` now uses `COALESCE(excluded.email, users.email)`. Two new tests.
- **Code review #3 — simplify `incrementHookUsage`** — now a single `INSERT ... ON CONFLICT ... RETURNING count` with no wrapping transaction and no defensive "row disappeared" branch. The dead `getHookUsage` prepared statement is removed. Existing concurrent-200 atomicity test still passes unchanged.
- **`setPlan` / Stripe note** — TODO comment added documenting the stripe_sub_id / renews_at wipe so #18 and #20 won't be surprised.

Coverage on `src/db/`: 95.21% lines / 100% functions / 88.57% branches (was 95.02 / 100 / 87.93). 111 tests pass.

PR #25, opened 2026-04-15
feat(oauth): bind message type as GCM AAD on delivery envelopes

dr14-make:
## Summary

Closes the domain-separation gap flagged in the OAuth broker design review: the ECIES envelope used for delivering OAuth tokens to the daemon now binds the envelope's `messageType` into the AES-256-GCM authenticated data on **both** ends of the flow.

Before: `eciesEncrypt` / `eciesDecrypt` only authenticated the ciphertext itself. The envelope's `Type` field travelled alongside the ciphertext but was not cryptographically bound to it, which in principle allowed a future attacker with access to the broker's encryption oracle to produce a ciphertext that a daemon would happily decrypt under a different `Type` label (cross-protocol attack on the same ECIES key).

After: both `eciesEncrypt` and `eciesDecrypt` require a non-empty `messageType` parameter and pass `Buffer.from(messageType, \"utf8\")` to `setAAD` before encrypt/decrypt. Any mismatch — including an in-transit tamper of the `Type` field — makes the GCM auth tag check fail, so the daemon identity key can never decrypt a ciphertext produced for a different message type.

`handleCallback` in `src/broker/router.ts` threads the literal `\"oauth_token_delivery\"` through as the AAD, which matches the type string the daemon side expects.

## Merge coordination

⚠ **Must land roughly together with `dicode-ayo/dicode-core#feat/oauth-relay-client`.** That PR adds the matching Go-side \`aead.Seal/Open\` AAD binding. Merging only one side will break OAuth token delivery: the encrypt and decrypt paths will disagree on the authenticated-data input and `aead.Open` will fail with an auth-tag error.

## Tests

- `tests/broker/crypto.test.ts`:
  - Round-trip: encrypt with `messageType=\"oauth_token_delivery\"`, decrypt with the same type, plaintext matches.
  - **Tampered type** — encrypt with one type, decrypt with a different type, expect an auth-tag failure.
  - **Empty type guard** — both encrypt and decrypt throw a clear \`domain separation\` error if `messageType` is empty, before any crypto runs.

## Test plan

- [ ] `npm test` green on CI
- [ ] Manual end-to-end with the coordinated dicode-core branch: run an OAuth flow against a real provider, confirm the daemon decrypts the delivery and writes the secret
- [ ] Negative: intercept the delivery POST and flip the `type` field, confirm the daemon rejects it with an auth-tag error
- [ ] Verify no other call site of `eciesEncrypt` / `eciesDecrypt` was missed (broker-side: `handleCallback` only)

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,6 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
-    "@iconify-icons/logos": "^1.2.36",
     "express": "^5.0.1",
     "grant": "^5.4.22",
     "js-yaml": "^4.1.1",
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
+    "@iconify-icons/logos": "^1.2.36",
     "express": "^5.0.1",
     "grant": "^5.4.22",
     "js-yaml": "^4.1.1",
--- a/src/broker/crypto.ts
+++ b/src/broker/crypto.ts
@@ -26,6 +26,16 @@ import { promisify } from "node:util";
 const hkdfAsync = promisify(hkdf);
 
 // ---------------------------------------------------------------------------
+// ECIES message type — ASCII-only by contract so that both
+// Buffer.from(type, "utf8") (TS) and []byte(type) (Go) produce
+// identical AAD bytes. A new envelope version must be introduced as a
+// new union member, not as a silent wire-format bump under the same label.
+// ---------------------------------------------------------------------------
+
+/** Discriminated type labels for ECIES-encrypted message envelopes. */
+export type EciesMessageType = "oauth_token_delivery";
+
+// ---------------------------------------------------------------------------
 // buildSignedPayload
 // ---------------------------------------------------------------------------
 
@@ -128,7 +138,7 @@ export interface EciesPayload {
 export async function eciesEncrypt(
   daemonPubkeyBytes: Buffer,
   sessionId: string,
-  messageType: string,
+  messageType: EciesMessageType,
   plaintext: Buffer,
 ): Promise<EciesPayload> {
   if (!messageType) {
@@ -174,14 +184,15 @@ export async function eciesEncrypt(
  * Decrypt an ECIES payload using the daemon's private key.
  * This mirrors what the Go daemon does on receipt of an oauth_token_delivery message.
  *
- * @param daemonPrivKeyPem PEM-encoded PKCS#8 private key
- * @param sessionId        Must match the salt used during encryption
- * @param payload          The EciesPayload from eciesEncrypt
+ * @param daemonECDH   ECDH instance holding the daemon's private key (prime256v1)
+ * @param sessionId    Must match the salt used during encryption
+ * @param messageType  Domain-separation label bound into GCM AAD (must match encrypt)
+ * @param payload      The EciesPayload from eciesEncrypt
  */
 export async function eciesDecrypt(
   daemonECDH: ReturnType<typeof createECDH>,
   sessionId: string,
-  messageType: string,
+  messageType: EciesMessageType,
   payload: EciesPayload,
 ): Promise<Buffer> {
   if (!messageType) {
--- a/src/broker/crypto.ts
+++ b/src/broker/crypto.ts
@@ -118,14 +118,22 @@ export interface EciesPayload {
  *   1. Generate ephemeral P-256 keypair
  *   2. ECDH(ephemeral_priv, daemon_pub) → shared_secret
  *   3. HKDF-SHA256(shared_secret, salt=session_id, info="dicode-oauth-token") → 32-byte enc_key
- *   4. AES-256-GCM encrypt with random 12-byte nonce
+ *   4. AES-256-GCM encrypt with random 12-byte nonce, binding `messageType`
+ *      into the GCM authenticated data so the envelope's Type field cannot
+ *      be swapped for a different message type without invalidating the
+ *      auth tag. This is domain separation: the same key is never asked
+ *      to decrypt two semantically distinct wire formats cleanly.
  *   5. Append 16-byte auth tag to ciphertext
  */
 export async function eciesEncrypt(
   daemonPubkeyBytes: Buffer,
   sessionId: string,
+  messageType: string,
   plaintext: Buffer,
 ): Promise<EciesPayload> {
+  if (!messageType) {
+    throw new Error("eciesEncrypt: messageType is required (domain separation)");
+  }
   const eph = createECDH("prime256v1");
   eph.generateKeys();
 
@@ -144,6 +152,7 @@ export async function eciesEncrypt(
 
   const iv = randomBytes(12);
   const cipher = createCipheriv("aes-256-gcm", encKey, iv);
+  cipher.setAAD(Buffer.from(messageType, "utf8"));
   const ctWithoutTag = Buffer.concat([cipher.update(plaintext), cipher.final()]);
   const authTag = cipher.getAuthTag(); // 16 bytes
 
@@ -172,8 +181,12 @@ export async function eciesEncrypt(
 export async function eciesDecrypt(
   daemonECDH: ReturnType<typeof createECDH>,
   sessionId: string,
+  messageType: string,
   payload: EciesPayload,
 ): Promise<Buffer> {
+  if (!messageType) {
+    throw new Error("eciesDecrypt: messageType is required (domain separation)");
+  }
   const ephPubBytes = Buffer.from(payload.ephemeralPubkey, "base64");
   const sharedSecret = daemonECDH.computeSecret(ephPubBytes);
 
@@ -194,6 +207,7 @@ export async function eciesDecrypt(
   const authTag = ciphertextWithTag.subarray(ciphertextWithTag.length - 16);
 
   const decipher = createDecipheriv("aes-256-gcm", encKey, iv);
+  decipher.setAAD(Buffer.from(messageType, "utf8"));
   decipher.setAuthTag(authTag);
 
   return Buffer.concat([decipher.update(ct), decipher.final()]);
--- a/src/broker/crypto.ts
+++ b/src/broker/crypto.ts
@@ -141,9 +141,7 @@ export async function eciesEncrypt(
   messageType: EciesMessageType,
   plaintext: Buffer,
 ): Promise<EciesPayload> {
-  if (!messageType) {
-    throw new Error("eciesEncrypt: messageType is required (domain separation)");
-  }
+  // messageType is type-enforced via EciesMessageType union — always truthy.
   const eph = createECDH("prime256v1");
   eph.generateKeys();
 
@@ -195,9 +193,6 @@ export async function eciesDecrypt(
   messageType: EciesMessageType,
   payload: EciesPayload,
 ): Promise<Buffer> {
-  if (!messageType) {
-    throw new Error("eciesDecrypt: messageType is required (domain separation)");
-  }
   const ephPubBytes = Buffer.from(payload.ephemeralPubkey, "base64");
   const sharedSecret = daemonECDH.computeSecret(ephPubBytes);
 
--- a/src/broker/router.ts
+++ b/src/broker/router.ts
@@ -197,18 +197,21 @@ async function handleCallback(
     Object.entries(rawTokenData).filter(([k]) => k !== "state"),
   );
 
-  // Encrypt the token payload with ECIES for the daemon
+  // Encrypt the token payload with ECIES for the daemon. The message type
+  // is bound into GCM's authenticated data so the daemon can never decrypt
+  // this envelope under a different type label — see crypto.ts.
+  const deliveryType = "oauth_token_delivery";
   const plaintext = Buffer.from(JSON.stringify(tokensToDeliver));
   let encrypted: Awaited<ReturnType<typeof eciesEncrypt>>;
   try {
-    encrypted = await eciesEncrypt(session.pubkey, session.sessionId, plaintext);
+    encrypted = await eciesEncrypt(session.pubkey, session.sessionId, deliveryType, plaintext);
   } catch {
     res.status(500).send("<html><body><p>Encryption failed</p></body></html>");
     return;
   }
 
   const deliveryPayload: OAuthTokenDeliveryPayload = {
-    type: "oauth_token_delivery",
+    type: deliveryType,
     session_id: session.sessionId,
     ephemeral_pubkey: encrypted.ephemeralPubkey,
     ciphertext: encrypted.ciphertext,
--- a/src/broker/router.ts
+++ b/src/broker/router.ts
@@ -17,20 +17,23 @@ import { buildSignedPayload, eciesEncrypt, verifyECDSA } from "./crypto.js";
 import type { ProviderConfig } from "./providers.js";
 import type { SessionStore } from "./sessions.js";
 import type { OAuthTokenDeliveryPayload } from "../relay/protocol.js";
+import { buildDeliverySignaturePayload, type BrokerSigningKey } from "./signing.js";
 
 const TIMESTAMP_TOLERANCE_S = 30;
 
 /**
  * Build the Express router for the OAuth broker.
  *
- * @param relay     RelayServer instance — used to look up clients and forward tokens
- * @param sessions  SessionStore instance
- * @param providers Enabled provider map (from config, already resolved)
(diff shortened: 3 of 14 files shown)


dr14-make (commented):
## Code Review

Nice, tight change. The fix is localised, the tests cover both directions of the AAD binding, and the guard fails fast before any crypto state is touched. A few non-blocking notes below.

**Scope creep in `package.json`**

The diff adds `@iconify-icons/logos@^1.2.36` (and pulls in `@iconify/types` transitively) to `dependencies`. That has nothing to do with OAuth AAD domain separation and I couldn't see anywhere it is imported from `src/`. It looks like an unrelated local install got swept into the commit. Please pull it back out of this PR so the change stays reviewable as a pure crypto fix — if it is needed for something else, it deserves its own PR.

**Type safety at call sites**

`messageType: string` is correct but wide. Every current and foreseeable caller will pass a fixed literal (`"oauth_token_delivery"`, and in future maybe one or two more). Consider narrowing to a union — `type EciesMessageType = "oauth_token_delivery";` — so a typo at a call site becomes a compile error instead of a runtime auth-tag failure at the daemon. The Go side already has an explicit allowlist for exactly this reason; mirroring that discipline in TS is cheap.

**Guard semantics**

`if (!messageType)` rejects `""` and `undefined`, but silently accepts `" "`, `"undefined"`, or any other junk string. That's fine as a last-ditch safety net, but combined with the literal-union suggestion above it becomes belt-and-braces. At minimum worth a one-line comment clarifying that the guard is defence-in-depth, not validation.

**Error messages**

`"eciesEncrypt: messageType is required (domain separation)"` is fine — it leaks nothing. Good. `handleCallback` wraps the call in `try { ... } catch { res.status(500).send("Encryption failed") }`, which is also appropriately opaque to the HTTP client. Good.

**Test coverage**

- Round-trip updated to pass `"oauth_token_delivery"` on both sides. OK.
- `"decryption fails with wrong message type"` — this is the right test: it encrypts with one type and decrypts with a *different* non-empty type, so the failure comes from `decipher.final()` (the AAD mismatch in the AEAD), not from the guard. Exactly what a domain-separation test should prove.
- `"encryption rejects empty message type"` — covers the guard path.
- Missing: a symmetric `eciesDecrypt` empty-type guard test. Trivial to add.
- Missing: a test that actually cross-checks byte-for-byte compatibility with a known Go-produced ciphertext (or vice versa). The TS test suite can't catch a future divergence between `Buffer.from(type, "utf8")` in TS and `[]byte(type)` in Go. For `"oauth_token_delivery"` (pure ASCII) they're identical, but a fixture-based interop test would lock that in.

**Stale JSDoc**

In `src/broker/crypto.ts` around the `eciesDecrypt` doc block, `@param daemonPrivKeyPem PEM-encoded PKCS#8 private key` no longer matches the actual signature (which takes `daemonECDH`). Pre-existing, but you're editing this block anyway — easy to fix in passing.

**Coordination check**

Confirmed against `dicode-ayo/dicode-core` `pkg/relay/oauth.go`: the Go side binds `[]byte(payload.Type)` as AAD and — critically — validates `payload.Type == "oauth_token_delivery"` *before* using it, so the chicken-and-egg concern about using envelope-provided bytes as their own AAD is neutralised. Good. Merging these two PRs out of order will break delivery, which is expected and should be called out in the merge plan.

package.json
```diff
@@ -21,6 +21,7 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
+    "@iconify-icons/logos": "^1.2.36",
```
dr14-make: Unrelated dependency addition. `@iconify-icons/logos` isn't used anywhere in `src/` and is unrelated to the OAuth AAD domain-separation fix. Please remove it from this PR and land it separately if it's needed for something else.

src/broker/crypto.ts
```diff
@@ -118,14 +118,22 @@ export interface EciesPayload {
...
+ *      into the GCM authenticated data so the envelope's Type field cannot
+ *      be swapped for a different message type without invalidating the
+ *      auth tag. This is domain separation: the same key is never asked
+ *      to decrypt two semantically distinct wire formats cleanly.
  *   5. Append 16-byte auth tag to ciphertext
  */
 export async function eciesEncrypt(
   daemonPubkeyBytes: Buffer,
   sessionId: string,
+  messageType: string,
```
dr14-make: Consider narrowing this to a literal union: `messageType: "oauth_token_delivery"` (or a `type EciesMessageType = ...` alias you extend as new types are added). A `string` will cheerfully accept typos at call sites, which then fail as auth-tag mismatches on the Go side — much harder to diagnose than a compile error. The Go daemon has an explicit allowlist for the same reason; mirroring that here keeps both sides honest.

src/broker/crypto.ts
```diff
@@ -118,14 +118,22 @@ export interface EciesPayload {
...
   plaintext: Buffer,
 ): Promise<EciesPayload> {
+  if (!messageType) {
+    throw new Error("eciesEncrypt: messageType is required (domain separation)");
```
dr14-make: Guard placement is correct — fails before any ECDH / HKDF work. Nit: `!messageType` also accepts `" "`, `"undefined"`, `"null"`, etc. Fine as defence-in-depth, but worth a short comment noting that real validation lives at the type system (and on the daemon side), and this is just a belt for the braces.

src/broker/crypto.ts
```diff
@@ -172,8 +181,12 @@ export async function eciesEncrypt(
 export async function eciesDecrypt(
   daemonECDH: ReturnType<typeof createECDH>,
   sessionId: string,
+  messageType: string,
   payload: EciesPayload,
 ): Promise<Buffer> {
+  if (!messageType) {
+    throw new Error("eciesDecrypt: messageType is required (domain separation)");
```
dr14-make: Consider adding a symmetric unit test for the decrypt-side empty-type guard; currently only the encrypt-side guard is covered. Very small, fully mechanical.

tests/broker/crypto.test.ts
```diff
@@ -247,12 +247,41 @@ describe("ECIES round-trip", () => {
...
+      daemonPubkeyBytes,
+      sessionId,
+      "oauth_token_delivery",
+      Buffer.from("t"),
+    );
+
+    // Decrypting with any other type label must fail.
+    await expect(
+      eciesDecrypt(daemonECDH, sessionId, "some_future_message", payload),
+    ).rejects.toThrow();
```
dr14-make, line 302: This is the right test — the failure here must come from `decipher.final()` (AEAD rejecting the AAD mismatch), not from the `!messageType` guard, because `"some_future_message"` is non-empty. Good coverage of the actual domain-separation guarantee.

One addition worth considering: a fixture-based cross-language interop test. The whole point of this PR is that Go's `[]byte("oauth_token_delivery")` and TS's `Buffer.from("oauth_token_delivery", "utf8")` produce byte-identical AAD. For pure ASCII that's trivially true today, but a frozen ciphertext fixture (produced once from the Go side) that the TS suite decrypts would catch any future encoding drift.

src/broker/router.ts
```diff
@@ -192,18 +192,21 @@ async function handleCallback(
     Object.entries(rawTokenData).filter(([k]) => k !== "state"),
   );
 
-  // Encrypt the token payload with ECIES for the daemon
+  // Encrypt the token payload with ECIES for the daemon. The message type
+  // is bound into GCM's authenticated data so the daemon can never decrypt
+  // this envelope under a different type label — see crypto.ts.
+  const deliveryType = "oauth_token_delivery";
```
dr14-make, line 213: Reasonable to pull this into a local `const` and reuse it for both the AAD and the envelope's `type` field — guarantees they can't drift. Once `eciesEncrypt`'s parameter is a literal union, this becomes a named constant on the type itself and you can drop the local binding if you want.

src/broker/crypto.ts
```diff
@@ -144,6 +152,7 @@ export async function eciesEncrypt(
 
   const iv = randomBytes(12);
   const cipher = createCipheriv("aes-256-gcm", encKey, iv);
+  cipher.setAAD(Buffer.from(messageType, "utf8"));
```
dr14-make, line 163: Confirmed byte-identical to the Go side's `[]byte(payload.Type)` for `"oauth_token_delivery"` (pure ASCII, 20 bytes). The chicken-and-egg concern about using envelope-provided bytes as their own AAD is neutralised on the daemon side by an explicit `payload.Type != "oauth_token_delivery"` allowlist check **before** `aead.Open` is called — so the AEAD is effectively being fed a fixed constant, just routed through the envelope field.

Suggest a short comment here pointing readers at the Go-side allowlist, since it's the thing that makes this construction safe and it's not locally obvious.

src/broker/crypto.ts
```diff
@@ -118,14 +118,22 @@ export interface EciesPayload {
...
+ *      to decrypt two semantically distinct wire formats cleanly.
  *   5. Append 16-byte auth tag to ciphertext
  */
 export async function eciesEncrypt(
   daemonPubkeyBytes: Buffer,
   sessionId: string,
+  messageType: string,
   plaintext: Buffer,
 ): Promise<EciesPayload> {
+  if (!messageType) {
```
dr14-make: Not a bug today because the only caller is `handleCallback` with a hardcoded literal, but: `!messageType` accepts `"undefined"`, `"null"`, `" "`, etc. If a future caller ever takes `messageType` from untrusted input, this guard is not enough. The TypeScript-level fix (literal union) closes this properly.

src/broker/crypto.ts
```diff
@@ -144,6 +152,7 @@ export async function eciesEncrypt(
 
   const iv = randomBytes(12);
```
dr14-make, line 161: Nonce construction unchanged by this PR — noting for the record that `randomBytes(12)` is called fresh per encryption, and each call also uses a fresh HKDF-derived key (fresh ephemeral ECDH), so effective IV reuse risk is zero. Good.

dr14-make (commented):
## Security Review

Adversarial pass. TL;DR: the fix does what it says on the tin, the Go-side coordination is sound, and I don't see an exploitable gap introduced by this PR. A few observations and one latent concern worth tracking.

**AAD correctness and chicken-and-egg**

The obvious worry with "bind `type` as AAD" when `type` travels inside the envelope is that the decrypt side has to trust envelope-provided bytes to validate envelope-provided bytes. I checked the Go side (`dicode-ayo/dicode-core` `pkg/relay/oauth.go`, function `DecryptOAuthToken`) and it does the right thing: it **rejects any `payload.Type != "oauth_token_delivery"` before calling `aead.Open`**, then passes `[]byte(payload.Type)` as AAD. So in effect the daemon is binding the *fixed string* `"oauth_token_delivery"` as AAD, it just routes through the envelope field. That collapses the concern — a tampered-in-transit `Type` field never reaches `aead.Open`, and an unknown `Type` never reaches `aead.Open` either. Good.

**Byte-identical AAD across languages**

- TS: `Buffer.from("oauth_token_delivery", "utf8")` → 20 bytes, `6f 61 75 74 68 5f 74 6f 6b 65 6e 5f 64 65 6c 69 76 65 72 79`.
- Go: `[]byte("oauth_token_delivery")` → identical 20 bytes.
- No BOM, no CRLF, no surrogate weirdness.

These agree today, trivially. They will agree tomorrow for any ASCII-only type label. If a future type label ever contains non-ASCII, both sides must be audited again — not a bug here, just a discipline reminder. I'd personally keep a grep-friendly comment saying "ASCII-only by contract" at both binding sites.

**Nonce / IV reuse**

Unchanged by this PR, but verified in passing: `eciesEncrypt` calls `randomBytes(12)` fresh on every invocation, and the "each encryption produces different ciphertext" test still holds. 12-byte random IV is the standard GCM construction with 2^32 messages as a safe bound under a fixed key — and every invocation here uses a fresh HKDF-derived key anyway (fresh ephemeral ECDH per message), so effective IV reuse risk is zero. Good.

**Empty-type guard**

`!messageType` correctly rejects `""` and `undefined`. It does *not* reject `"undefined"`, `"null"`, `" "`, or other pathological strings that an attacker who controls a caller could potentially supply. This is only a concern if an attacker can control `messageType` at a call site — which they can't, because the only caller is `handleCallback` with the hardcoded literal `"oauth_token_delivery"`. So: not exploitable today, but a future callsite that takes `messageType` from user-controlled input would be. Tracking via the code-review suggestion to narrow the type to a literal union.

**`handleCallback` failure modes**

Encryption happens **after** state validation (state + access_token present, session lookup, session exists). An attacker cannot reach `eciesEncrypt` without producing a valid `state` that hits a live session. Session is deleted immediately after encryption succeeds. The `catch { res.status(500).send("Encryption failed") }` is generic and leaks nothing. No path I can see where a crafted callback could make `eciesEncrypt` crash in a way that exposes key material — the key is ephemeral and lives only on the stack for this one call.

One subtle thing: on encryption failure the session is *not* deleted (delete is after the try/catch). An attacker who can repeatedly trigger encryption failures could keep a session alive past its expected single-use lifetime. Extremely unlikely to be exploitable (the cause of the failure is essentially "Node crypto broke"), but worth noting that the single-use invariant has a small hole on the error path. Not introduced by this PR.

**Tamper testing on the decrypt side**

The new TS test `"decryption fails with wrong message type"` encrypts with `"oauth_token_delivery"` and decrypts with `"some_future_message"`. Both are non-empty, so the failure is not the `!messageType` guard firing — it's `decipher.final()` throwing on the AEAD tag mismatch. That's exactly the right shape for an AAD test. The Go side has a symmetric test in `pkg/relay/oauth_test.go` (`TestDecryptOAuthToken_RejectsTamperedType`). Coverage-wise this is solid.

Missing from the TS suite: a test that reproduces the *wire* attack — hand-craft an envelope with ciphertext from a `"oauth_token_delivery"` session but `type: "oauth_refresh_delivery"` in the transport JSON — and confirm the daemon rejects it. That can only be tested on the Go side (since TS only has the encrypt side in production), and the Go PR should cover it. Worth confirming.

**Future-proofing: no version byte**

The envelope has no `version` field. If `crypto.ts` ever ships a v2 envelope (different KDF info string, different AAD construction, different nonce size, whatever), there is no way to tell v1 from v2 on the wire without adding a new type label. That's *fine* for this PR — in fact it's the whole point of domain separation via `type` — but it means v2 needs to ship as a new `type` literal (e.g. `"oauth_token_delivery_v2"`), not as a silent wire-format bump under the same type. Please make sure this expectation is captured somewhere the next engineer will see it (a comment in `crypto.ts` above the AAD binding, or in `docs/oauth.md`). It's a subtle rule and easy to violate by accident.

**No findings that block merge.** The domain-separation fix is correct, consistent with the Go side, and well-tested. Remove the unrelated `@iconify-icons/logos` dep (see code review) and ship it with the matching core PR.

src/broker/router.ts
```diff
@@ -192,18 +192,21 @@ async function handleCallback(
...
   const plaintext = Buffer.from(JSON.stringify(tokensToDeliver));
   let encrypted: Awaited<ReturnType<typeof eciesEncrypt>>;
   try {
-    encrypted = await eciesEncrypt(session.pubkey, session.sessionId, plaintext);
+    encrypted = await eciesEncrypt(session.pubkey, session.sessionId, deliveryType, plaintext);
```
dr14-make, line 217: Encryption is correctly placed *after* all validation (state present, session found). Minor observation unrelated to this PR: on the `catch` branch, `sessions.delete(session.sessionId)` has not yet run, so an attacker who can repeatedly trigger `eciesEncrypt` failure keeps the session alive past its intended single-use window. Not exploitable today (the failure mode is essentially "Node crypto broke"), but worth tracking if the single-use invariant ever gets relied on as a hard security property.

tests/broker/crypto.test.ts
```diff
@@ -247,12 +247,41 @@ describe("ECIES round-trip", () => {
...
+    const payload = await eciesEncrypt(
+      daemonPubkeyBytes,
+      sessionId,
+      "oauth_token_delivery",
+      Buffer.from("t"),
+    );
+
+    // Decrypting with any other type label must fail.
+    await expect(
+      eciesDecrypt(daemonECDH, sessionId, "some_future_message", payload),
```
dr14-make: Correct shape: `"some_future_message"` is non-empty, so the failure path exercised here is the AEAD tag rejecting a mismatched AAD in `decipher.final()`, not the `!messageType` guard. This is what a true AAD test should look like.

The symmetric case — hand-crafting an envelope with a valid ciphertext but a swapped `type` field on the wire — can only be tested on the Go side (the TS codebase doesn't decrypt in production). Worth confirming the matching core PR has that test.

PR #32, opened 2026-04-22
feat(broker): E2E mock OAuth provider gated by DICODE_E2E_MOCK_PROVIDER

### LINKED ISSUE #31: feat: mock OAuth provider for E2E (gated by DICODE_E2E_MOCK_PROVIDER)
Part of dicode-core#136 (cross-service E2E test epic).

## Problem

The real OAuth broker flow requires registered \`client_id\` / \`client_secret\` values for each upstream provider (GitHub, Slack, etc.). Using real credentials in CI is:
- Fragile (rate limits, revocations)
- Leaky (secrets in CI configuration)
- Slow (browser redirects, consent screens)

For dicode-core's E2E suite to exercise the full token-delivery path (broker signs → ECIES encrypts → daemon receives → daemon decrypts), we need a **fake provider** that short-circuits the authorize + token-exchange steps but preserves everything the daemon sees.

## Proposed design

Gate behind a single env var so it cannot leak into production by accident:

\`\`\`
DICODE_E2E_MOCK_PROVIDER=1
\`\`\`

When set, register a new provider \`mock\` alongside the real ones. Routes:

- \`GET /connect/mock?state=<session>\` — immediately redirects to \`GET /callback/mock?code=mock-code-<session>&state=<session>\` (skips the upstream redirect)
- \`GET /callback/mock\` — Grant still processes this; we shim it so the \"token exchange\" resolves to \`{access_token: \"mock-token-<session>\", token_type: \"bearer\"}\` without any outbound HTTP

Implementation options:
1. Grant adapter with custom \`authorize_url\` + \`access_url\` pointing at loopback endpoints served in the same process
2. Bypass Grant entirely for \`mock\` — directly insert a fake token into the session and trigger the existing ECIES encrypt + forward path

Option 2 is simpler and keeps the mock isolated from Grant's internals.

## Acceptance

- [ ] \`mock\` provider registers only when \`DICODE_E2E_MOCK_PROVIDER=1\`
- [ ] Daemon's OAuth IPC (\`dicode.oauth.build_auth_url\`) returns a URL pointing at the broker's \`/auth/mock\` route
- [ ] Opening that URL (via the daemon's webhook → browser flow) round-trips to an ECIES-encrypted token delivery to the daemon
- [ ] Daemon's existing \`DecryptOAuthToken\` verifies the delivered token equals \`mock-token-<sessionId>\`
- [ ] Startup log prints a bright WARN when the flag is on: \"DICODE_E2E_MOCK_PROVIDER enabled — DO NOT RUN IN PRODUCTION\"
- [ ] CI test validates the flag is **off** in production builds (grep the docker image for the env default)

## Out of scope

- Mocking real provider quirks (Notion's Basic auth, Stripe's \`stripe_user_id\`) — the mock is for **wire-path** testing, not per-provider edge cases
- Per-test scope / claim customization — the mock returns a fixed payload; if a test needs specific claims it can configure them via additional env vars in a follow-up

dr14-make:
## Summary

Closes #31. Adds a mock OAuth provider that short-circuits the upstream authorize + token-exchange steps, so CI / integration tests can exercise the full daemon-issued `build_auth_url` → broker → daemon token-delivery round-trip without any real OAuth credentials.

Two endpoints, both gated behind `DICODE_E2E_MOCK_PROVIDER=1` (a WARN is logged at startup, same mount-before-Grant pattern the issue spec describes):

- **`GET /connect/mock`** — reads the session created by `/auth/mock`, redirects the browser to `/callback/mock?state=…&access_token=mock-token-<session>&token_type=bearer`. The existing `/callback/:provider` handler then encrypts + signs + forwards to the daemon. This is the flow acceptance criteria #2–#4 describe.
- **`POST /_test/deliver`** — lower-level primitive that bypasses `/auth` and `/connect` entirely. Takes `{uuid, session_id, provider, tokens}` directly and exercises the encrypt + sign + forward path. This is the endpoint that uncovered dicode-core#151 (broker-sig hash-depth mismatch) and is worth keeping around for cross-implementation wire-shape testing.

`mock` is registered in `brokerProviders` so `/auth/mock` passes the provider-exists check, but deliberately excluded from the Grant config (option 2 in the issue spec) — Grant never tries to dispatch `/connect/mock` upstream.

## Acceptance-criteria coverage (from #31)

- [x] `mock` provider registers only when `DICODE_E2E_MOCK_PROVIDER=1`
- [x] Daemon's OAuth IPC → `/auth/mock` is accepted (provider map gating)
- [x] Opening the daemon-issued URL round-trips to an ECIES-encrypted token delivery
- [x] Daemon's `DecryptOAuthToken` sees `mock-token-<sessionId>` (validated E2E against a real dicode-core daemon when the broker-sig fix in dicode-core#152 landed)
- [x] Startup prints bright WARN when the flag is on
- [ ] Docker image asserts flag is OFF by default — left as a follow-up; worth a CI step in `.github/workflows/ci.yml` once the test suite covers more docker-related checks

## Test plan

- [x] `npm run typecheck && npm run lint && npm run format:check && npm run test` — all green (110/110)
- [x] New test file `tests/broker/e2e-mock.test.ts` covers:
  - `/connect/mock` valid session → 302 with synthetic token
  - `/connect/mock` missing / unknown / non-mock-provider state → 400
  - `/_test/deliver` missing fields → 400
  - `/_test/deliver` unknown daemon uuid → 404
  - `/_test/deliver` connected daemon → encrypts + signs + forwards; the delivery sig verifies against `brokerKey.publicKeyBase64`
- [x] Manually exercised against a live dicode-core daemon during dicode-core#151/#152 investigation — secrets (`GITHUB_ACCESS_TOKEN`, `GITHUB_REFRESH_TOKEN`, `GITHUB_SCOPE`, `GITHUB_TOKEN_TYPE`) correctly written on the daemon side after the fix

### DIFF
--- a/src/broker/e2e-mock.ts
+++ b/src/broker/e2e-mock.ts
@@ -43,9 +43,14 @@ export const MOCK_PROVIDER_KEY = "mock";
  * enabled flag in prod is "attacker writes chosen OAuth tokens into any
  * connected daemon's secret store." The NODE_ENV refusal converts the
  * operator-must-remember invariant into an enforced one.
+ *
+ * The NODE_ENV match is case-insensitive and whitespace-tolerant so that
+ * oddly-capitalised or copy-pasted values like "PRODUCTION" or
+ * " production" still trip the refusal.
  */
 export function isE2EMockEnabled(): boolean {
-  if (process.env.NODE_ENV === "production") return false;
+  const nodeEnv = process.env.NODE_ENV?.trim().toLowerCase();
+  if (nodeEnv === "production") return false;
   return process.env.DICODE_E2E_MOCK_PROVIDER === "1";
 }
 
--- a/src/broker/e2e-mock.ts
+++ b/src/broker/e2e-mock.ts
@@ -1,0 +1,169 @@
+/**
+ * E2E mock OAuth provider — gated behind DICODE_E2E_MOCK_PROVIDER=1.
+ * Must never mount in production.
+ *
+ * Exposes two endpoints, both useful for exercising the broker→daemon token
+ * delivery path end-to-end without any real upstream OAuth provider:
+ *
+ *   GET /connect/mock
+ *     Short-circuits the upstream authorize + code-exchange steps. Reads
+ *     `state` (the session ID created by /auth/mock), looks up the session,
+ *     and redirects the browser to /callback/mock?state=…&access_token=…
+ *     The existing /callback/:provider handler then encrypts and forwards
+ *     the synthetic token to the daemon. This is the flow dicode-relay#31
+ *     specifies, and it lets a test driver exercise the full daemon-issued
+ *     build_auth_url → browser → broker → daemon round-trip.
+ *
+ *   POST /_test/deliver
+ *     A lower-level primitive that bypasses /auth and /connect entirely.
+ *     Takes { uuid, session_id, provider, tokens } directly, builds the
+ *     ECIES envelope using the connected daemon's decrypt pubkey + the
+ *     broker signing key, and forwards. Used for cross-implementation
+ *     wire-shape testing (this is how dicode-core#151 — the broker-sig
+ *     hash-depth mismatch — was uncovered).
+ */
+
+import type { Request, Response, Router } from "express";
+import { Router as makeRouter, json } from "express";
+import { eciesEncrypt } from "./crypto.js";
+import type { OAuthTokenDeliveryPayload } from "../relay/protocol.js";
+import type { RelayServer } from "../relay/server.js";
+import type { SessionStore } from "./sessions.js";
+import { buildDeliverySignaturePayload, type BrokerSigningKey } from "./signing.js";
+
+/** Provider key used throughout the mock flow. */
+export const MOCK_PROVIDER_KEY = "mock";
+
+/** Whether the E2E mock provider is enabled for this process. */
+export function isE2EMockEnabled(): boolean {
+  return process.env.DICODE_E2E_MOCK_PROVIDER === "1";
+}
+
+interface DeliverBody {
+  uuid: string;
+  session_id: string;
+  provider: string;
+  tokens: Record<string, unknown>;
+}
+
+/**
+ * Build the Express router for the E2E mock provider.
+ *
+ * Mount BEFORE the Grant middleware so /connect/mock is handled here and
+ * Grant never sees it, and BEFORE the broker router so /_test/deliver
+ * is reachable.
+ */
+export function buildE2EMockRouter(
+  relay: RelayServer,
+  sessions: SessionStore,
+  brokerKey: BrokerSigningKey,
+  baseUrl: string,
+): Router {
+  const router: Router = makeRouter();
+  router.use(json());
+
+  router.get("/connect/mock", (req: Request, res: Response) => {
+    handleConnectMock(req, res, sessions, baseUrl);
+  });
+
+  router.post("/_test/deliver", (req: Request, res: Response) => {
+    void handleDeliver(req, res, relay, brokerKey);
+  });
+
+  return router;
+}
+
+function handleConnectMock(
+  req: Request,
+  res: Response,
+  sessions: SessionStore,
+  baseUrl: string,
+): void {
+  const rawState = req.query.state;
+  const state = Array.isArray(rawState) ? rawState[0] : rawState;
+  if (typeof state !== "string" || state === "") {
+    res.status(400).send("missing state");
+    return;
+  }
+
+  const session = sessions.get(state);
+  if (session === undefined) {
+    res.status(400).send("session not found");
+    return;
+  }
+  if (session.provider !== MOCK_PROVIDER_KEY) {
+    res.status(400).send("session is not for mock provider");
+    return;
+  }
+
+  const callbackUrl =
+    `${baseUrl}/callback/${MOCK_PROVIDER_KEY}` +
+    `?state=${encodeURIComponent(state)}` +
+    `&access_token=${encodeURIComponent(`mock-token-${state}`)}` +
+    `&token_type=bearer`;
+  res.setHeader("Referrer-Policy", "no-referrer");
+  res.redirect(302, callbackUrl);
+}
+
+async function handleDeliver(
+  req: Request,
+  res: Response,
+  relay: RelayServer,
+  brokerKey: BrokerSigningKey,
+): Promise<void> {
+  const body = req.body as Partial<DeliverBody>;
+  if (!body.uuid || !body.session_id || !body.provider || typeof body.tokens !== "object") {
+    res.status(400).json({ error: "uuid, session_id, provider, tokens required" });
+    return;
+  }
+
+  if (!relay.hasClient(body.uuid)) {
+    res.status(404).json({ error: "daemon uuid not connected to relay" });
+    return;
+  }
+  const client = relay.getClient(body.uuid);
+
+  const deliveryType = "oauth_token_delivery";
+  const plaintext = Buffer.from(JSON.stringify(body.tokens));
+
+  let encrypted: Awaited<ReturnType<typeof eciesEncrypt>>;
+  try {
+    encrypted = await eciesEncrypt(client.decryptPubkey, body.session_id, deliveryType, plaintext);
+  } catch (e) {
+    res.status(500).json({ error: `encrypt failed: ${String(e)}` });
+    return;
+  }
+
+  const payload: OAuthTokenDeliveryPayload = {
+    type: deliveryType,
+    session_id: body.session_id,
+    ephemeral_pubkey: encrypted.ephemeralPubkey,
+    ciphertext: encrypted.ciphertext,
+    nonce: encrypted.nonce,
+  };
+
+  const sigPayload = buildDeliverySignaturePayload(
+    payload.type,
+    payload.session_id,
+    payload.ephemeral_pubkey,
+    payload.ciphertext,
+    payload.nonce,
+  );
+  payload.broker_sig = brokerKey.sign(sigPayload);
+
+  try {
+    const daemonResp = await relay.forward(
+      body.uuid,
+      "POST",
+      "/hooks/oauth-complete",
+      { "Content-Type": ["application/json"] },
+      Buffer.from(JSON.stringify(payload)),
+    );
+    res.status(200).json({
+      daemon_status: daemonResp.status,
+      daemon_body: Buffer.from(daemonResp.body, "base64").toString("utf8"),
+    });
+  } catch (e) {
+    res.status(502).json({ error: `forward failed: ${String(e)}` });
+  }
+}
--- a/src/broker/e2e-mock.ts
+++ b/src/broker/e2e-mock.ts
@@ -34,8 +34,18 @@ import { buildDeliverySignaturePayload, type BrokerSigningKey } from "./signing.
 /** Provider key used throughout the mock flow. */
 export const MOCK_PROVIDER_KEY = "mock";
 
-/** Whether the E2E mock provider is enabled for this process. */
+/**
+ * Whether the E2E mock provider is enabled for this process.
+ *
+ * Fails closed when `NODE_ENV=production` even if the flag is set — the
+ * mock endpoints hand out daemon-decryptable token deliveries to anyone
+ * who can reach the HTTP port, so the consequence of an accidentally
+ * enabled flag in prod is "attacker writes chosen OAuth tokens into any
+ * connected daemon's secret store." The NODE_ENV refusal converts the
+ * operator-must-remember invariant into an enforced one.
+ */
 export function isE2EMockEnabled(): boolean {
+  if (process.env.NODE_ENV === "production") return false;
   return process.env.DICODE_E2E_MOCK_PROVIDER === "1";
 }
 
@@ -57,13 +67,12 @@ export function buildE2EMockRouter(
   relay: RelayServer,
   sessions: SessionStore,
   brokerKey: BrokerSigningKey,
-  baseUrl: string,
 ): Router {
   const router: Router = makeRouter();
   router.use(json());
 
   router.get("/connect/mock", (req: Request, res: Response) => {
-    handleConnectMock(req, res, sessions, baseUrl);
+    handleConnectMock(req, res, sessions);
   });
 
   router.post("/_test/deliver", (req: Request, res: Response) => {
@@ -73,12 +82,7 @@ export function buildE2EMockRouter(
   return router;
 }
 
-function handleConnectMock(
-  req: Request,
-  res: Response,
-  sessions: SessionStore,
-  baseUrl: string,
-): void {
+function handleConnectMock(req: Request, res: Response, sessions: SessionStore): void {
   const rawState = req.query.state;
   const state = Array.isArray(rawState) ? rawState[0] : rawState;
   if (typeof state !== "string" || state === "") {
@@ -95,14 +99,22 @@ function handleConnectMock(
     res.status(400).send("session is not for mock provider");
     return;
   }
+  if (session.expiresAt <= Date.now()) {
+    res.status(400).send("session expired");
+    return;
+  }
 
-  const callbackUrl =
-    `${baseUrl}/callback/${MOCK_PROVIDER_KEY}` +
+  const callbackPath =
+    `/callback/${MOCK_PROVIDER_KEY}` +
     `?state=${encodeURIComponent(state)}` +
     `&access_token=${encodeURIComponent(`mock-token-${state}`)}` +
     `&token_type=bearer`;
   res.setHeader("Referrer-Policy", "no-referrer");
-  res.redirect(302, callbackUrl);
+  res.redirect(302, callbackPath);
+}
+
+function isPlainTokensObject(v: unknown): v is Record<string, unknown> {
+  return typeof v === "object" && v !== null && !Array.isArray(v);
 }
 
 async function handleDeliver(
@@ -112,7 +124,15 @@ async function handleDeliver(
   brokerKey: BrokerSigningKey,
 ): Promise<void> {
   const body = req.body as Partial<DeliverBody>;
-  if (!body.uuid || !body.session_id || !body.provider || typeof body.tokens !== "object") {
+  if (
+    typeof body.uuid !== "string" ||
+    body.uuid === "" ||
+    typeof body.session_id !== "string" ||
+    body.session_id === "" ||
+    typeof body.provider !== "string" ||
+    body.provider === "" ||
+    !isPlainTokensObject(body.tokens)
+  ) {
     res.status(400).json({ error: "uuid, session_id, provider, tokens required" });
     return;
   }
@@ -130,7 +150,8 @@ async function handleDeliver(
   try {
     encrypted = await eciesEncrypt(client.decryptPubkey, body.session_id, deliveryType, plaintext);
   } catch (e) {
-    res.status(500).json({ error: `encrypt failed: ${String(e)}` });
+    console.error("e2e-mock: encrypt failed", e);
+    res.status(500).json({ error: "encrypt failed" });
     return;
   }
 
@@ -159,11 +180,12 @@ async function handleDeliver(
       { "Content-Type": ["application/json"] },
       Buffer.from(JSON.stringify(payload)),
     );
-    res.status(200).json({
-      daemon_status: daemonResp.status,
-      daemon_body: Buffer.from(daemonResp.body, "base64").toString("utf8"),
-    });
+    // Return only the daemon's HTTP status. The body is intentionally NOT
+    // reflected back — the caller already knows what they sent and the
+    // daemon's reply may contain noisy or sensitive error context.
+    res.status(200).json({ daemon_status: daemonResp.status });
   } catch (e) {
-    res.status(502).json({ error: `forward failed: ${String(e)}` });
+    console.error("e2e-mock: forward failed", e);
+    res.status(502).json({ error: "forward failed" });
   }
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -19,7 +19,8 @@ import { loadConfig } from "./config.js";
 import { RelayServer } from "./relay/server.js";
 import { buildGrantMiddleware } from "./broker/grant.js";
 import { buildBrokerRouter } from "./broker/router.js";
-import { buildProviderMap } from "./broker/providers.js";
+import { MOCK_PROVIDER_KEY, buildE2EMockRouter, isE2EMockEnabled } from "./broker/e2e-mock.js";
+import { buildProviderMap, type ProviderConfig } from "./broker/providers.js";
 import { SessionStore } from "./broker/sessions.js";
 import { loadBrokerSigningKey } from "./broker/signing.js";
 import { MetricsCollector } from "./status/metrics.js";
@@ -91,13 +92,39 @@ relayServer.on("client:disconnected", (uuid: string) => {
 // OAuth broker
 // ---------------------------------------------------------------------------
 
-const providers = buildProviderMap(config);
+const realProviders = buildProviderMap(config);
 const sessions = new SessionStore(brokerCfg.session_ttl_ms);
 
-const grantMiddleware = buildGrantMiddleware(providers, serverCfg.base_url);
+// Providers passed to the broker router (session creation in /auth/:provider).
+// If the E2E mock flag is set, include "mock" here so /auth/mock is accepted.
+// Grant must NOT receive the mock entry — /connect/mock is handled by the
+// e2e-mock router, and Grant would otherwise try to dispatch it upstream.
+const brokerProviders = new Map<string, ProviderConfig>(realProviders);
+if (isE2EMockEnabled()) {
+  brokerProviders.set(MOCK_PROVIDER_KEY, {
+    grantKey: MOCK_PROVIDER_KEY,
+    clientId: "mock-client-id",
+    pkce: true,
+    scopes: [],
+  });
+  console.warn(
+    "broker: DICODE_E2E_MOCK_PROVIDER enabled — mock provider registered. DO NOT USE IN PRODUCTION.",
+  );
+  // Mount BEFORE Grant so /connect/mock is intercepted and Grant never sees
+  // it. Also exposes /_test/deliver for low-level wire-shape testing.
+  app.use(buildE2EMockRouter(relayServer, sessions, brokerKey, serverCfg.base_url));
+}
+
+const grantMiddleware = buildGrantMiddleware(realProviders, serverCfg.base_url);
 app.use(grantMiddleware);
 app.use(
-  buildBrokerRouter(relayServer, sessions, providers, relayCfg.timestamp_tolerance_s, brokerKey),
+  buildBrokerRouter(
+    relayServer,
+    sessions,
+    brokerProviders,
+    relayCfg.timestamp_tolerance_s,
+    brokerKey,
+  ),
 );
 
 // ---------------------------------------------------------------------------
@@ -179,7 +206,7 @@ app.get("/health", (_req, res) => {
 httpServer.listen(serverCfg.port, () => {
   console.log(`dicode-relay listening on port ${String(serverCfg.port)}`);
   console.log(`Base URL: ${serverCfg.base_url}`);
-  console.log(`Providers: ${[...providers.keys()].join(", ") || "(none configured)"}`);
+  console.log(`Providers: ${[...brokerProviders.keys()].join(", ") || "(none configured)"}`);
 });
 
 // Graceful shutdown
--- a/src/index.ts
+++ b/src/index.ts
@@ -103,7 +103,10 @@ const brokerProviders = new Map<string, ProviderConfig>(realProviders);
 if (isE2EMockEnabled()) {
   brokerProviders.set(MOCK_PROVIDER_KEY, {
     grantKey: MOCK_PROVIDER_KEY,
-    clientId: "mock-client-id",
+    // Obviously-fake placeholder — never reaches Grant (see buildGrantMiddleware
+    // call below, which is passed realProviders only). Exists solely to satisfy
+    // the non-empty check in providers.has/buildProviderMap invariants.
+    clientId: "mock-e2e-not-a-real-credential",
     pkce: true,
     scopes: [],
   });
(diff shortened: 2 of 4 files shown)


src/broker/e2e-mock.ts
```diff
@@ -0,0 +1,169 @@
...
+}
+
+async function handleDeliver(
+  req: Request,
+  res: Response,
+  relay: RelayServer,
+  brokerKey: BrokerSigningKey,
+): Promise<void> {
+  const body = req.body as Partial<DeliverBody>;
+  if (!body.uuid || !body.session_id || !body.provider || typeof body.tokens !== "object") {
```
dr14-make: **suggestion** — `typeof body.tokens !== "object"` lets both `null` and arrays slip through the guard (`typeof null === "object"`, `typeof [] === "object"`). For the array case, `JSON.stringify` will happily emit `"[…]"` and the daemon will receive a non-object envelope. Suggest:

```ts
if (
  !body.uuid ||
  !body.session_id ||
  !body.provider ||
  body.tokens === null ||
  typeof body.tokens !== "object" ||
  Array.isArray(body.tokens)
) {
  res.status(400).json({ error: "uuid, session_id, provider, tokens required" });
  return;
}
```

Low-impact since this endpoint is test-only, but it's one line and keeps the wire-shape contract strict — which is exactly what `/_test/deliver` is for.

src/broker/e2e-mock.ts
```diff
@@ -0,0 +1,169 @@
...
+  const client = relay.getClient(body.uuid);
+
+  const deliveryType = "oauth_token_delivery";
+  const plaintext = Buffer.from(JSON.stringify(body.tokens));
+
+  let encrypted: Awaited<ReturnType<typeof eciesEncrypt>>;
+  try {
+    encrypted = await eciesEncrypt(client.decryptPubkey, body.session_id, deliveryType, plaintext);
+  } catch (e) {
+    res.status(500).json({ error: `encrypt failed: ${String(e)}` });
```
dr14-make: **nit** — ```res.status(500).json({ error: `encrypt failed: ${String(e)}` })``` (and the 502 `forward failed: …` below) leak raw error messages to the HTTP response. For a test-only endpoint that's fine, but it would be a footgun if anyone ever re-used `handleDeliver` from a prod path. Worth a comment above the function clarifying that the route is DICODE_E2E_MOCK_PROVIDER-gated for exactly this reason, and/or logging the error server-side and returning an opaque message.

src/broker/e2e-mock.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  const callbackUrl =
+    `${baseUrl}/callback/${MOCK_PROVIDER_KEY}` +
+    `?state=${encodeURIComponent(state)}` +
+    `&access_token=${encodeURIComponent(`mock-token-${state}`)}` +
+    `&token_type=bearer`;
+  res.setHeader("Referrer-Policy", "no-referrer");
+  res.redirect(302, callbackUrl);
+}
```
dr14-make, line 123: **nit** — the real broker router uses a *relative* redirect on its equivalent hop (`res.redirect(302, redirectPath)` in router.ts, L154) to avoid baking a cross-origin hop when the service is fronted by a reverse proxy that mismatches `base_url`. Here the redirect target is always same-origin (`/callback/mock`), so a relative path would work and match the pattern:

```ts
const callbackPath =
  `/callback/${MOCK_PROVIDER_KEY}` +
  `?state=${encodeURIComponent(state)}` +
  `&access_token=${encodeURIComponent(\`mock-token-${state}\`)}` +
  `&token_type=bearer`;
res.redirect(302, callbackPath);
```

Not a bug — full URL is fine — but it drops the `baseUrl` param dependency from the helper, removes the config-mismatch failure mode, and makes the mock route consistent with `router.ts`. You asked about this specifically (focus #3); this is my take.

src/index.ts
```diff
@@ -91,13 +92,39 @@ relayServer.on("client:disconnected", (uuid: string) => {
...
+    clientId: "mock-client-id",
+    pkce: true,
+    scopes: [],
+  });
+  console.warn(
+    "broker: DICODE_E2E_MOCK_PROVIDER enabled — mock provider registered. DO NOT USE IN PRODUCTION.",
+  );
+  // Mount BEFORE Grant so /connect/mock is intercepted and Grant never sees
+  // it. Also exposes /_test/deliver for low-level wire-shape testing.
+  app.use(buildE2EMockRouter(relayServer, sessions, brokerKey, serverCfg.base_url));
```
dr14-make: **suggestion** — `Dockerfile` doesn't set `DICODE_E2E_MOCK_PROVIDER=0` (or explicitly `unset` it) in the runtime stage. If the flag bleeds through from a docker run env or a compose file somewhere (`docker run -e DICODE_E2E_MOCK_PROVIDER=1`) the mock would be active in a prod image. You already flagged this as a follow-up in the PR description, but two belt-and-suspenders hardenings would close the loop cheaply:

1. Add `ENV DICODE_E2E_MOCK_PROVIDER=0` to the runtime stage of the Dockerfile so operators have to actively override.
2. When `isE2EMockEnabled()` returns true, also refuse to start unless `process.env.NODE_ENV !== "production"` — or log WARN + continue, your call. Current WARN is good but silent-ignored in systemd/k8s.

The first one alone is maybe 10 seconds and closes most of the accident-prone surface. Consider making this part of this PR rather than a follow-up.

src/broker/e2e-mock.ts
```diff
@@ -0,0 +1,169 @@
...
+import type { RelayServer } from "../relay/server.js";
+import type { SessionStore } from "./sessions.js";
+import { buildDeliverySignaturePayload, type BrokerSigningKey } from "./signing.js";
+
+/** Provider key used throughout the mock flow. */
+export const MOCK_PROVIDER_KEY = "mock";
+
+/** Whether the E2E mock provider is enabled for this process. */
+export function isE2EMockEnabled(): boolean {
+  return process.env.DICODE_E2E_MOCK_PROVIDER === "1";
```
dr14-make, line 54: **[high] Missing production safety net — no CI / Dockerfile assertion that `DICODE_E2E_MOCK_PROVIDER` is unset.**

The PR description lists "Docker image asserts flag is OFF by default" as an unchecked acceptance criterion from #31, deferred to a follow-up. This is the single largest risk in this PR: the entire security model (two unauthenticated endpoints that can write arbitrary OAuth tokens into any connected daemon's secret store) rests on this one env var being unset in production. There is nothing stopping an operator from accidentally setting `DICODE_E2E_MOCK_PROVIDER=1` in a prod deployment and leaving it that way.

Please add at least one of the following before merging:

1. **Dockerfile line** — explicitly set `ENV DICODE_E2E_MOCK_PROVIDER=""` in the runtime stage so that even if the build or compose layer tries to inherit a stray value, the final image forces it to empty.
2. **CI grep check** — a `.github/workflows/ci.yml` step that fails if `DICODE_E2E_MOCK_PROVIDER` appears set to `1` in any shipped artifact (Dockerfile, compose files, helm charts, env.example, etc.).
3. **Production-mode refusal** — if `NODE_ENV === "production"`, either ignore the flag entirely or abort startup with a loud error. This is the strongest defense and costs nothing at runtime.

Option 3 is the one I'd strongly recommend — it would make accidental enablement in production impossible rather than merely observable after the fact.

src/broker/e2e-mock.ts
```diff
@@ -0,0 +1,169 @@
...
+
+  const session = sessions.get(state);
+  if (session === undefined) {
+    res.status(400).send("session not found");
+    return;
+  }
+  if (session.provider !== MOCK_PROVIDER_KEY) {
+    res.status(400).send("session is not for mock provider");
+    return;
+  }
```
dr14-make, line 106: **suggestion** — you asked (focus #7) about which session invariants to guard. Currently the flow is:
- provider === "mock" ✓ checked
- session exists ✓ checked (via `sessions.get`)
- session not expired — NOT explicitly checked. `SessionStore.get` does not enforce TTL on read; it relies on a setTimeout to delete. That's fine for the 5-min TTL window, but if the setTimeout fires late under event-loop pressure, an expired session could still redirect to `/callback/mock`. The downstream `/callback/:provider` handler also doesn't check `expiresAt`, so this isn't mock-specific — but since you asked, making the mock flow tighter than the real flow would be cheap:

```ts
if (Date.now() > session.expiresAt) {
  res.status(400).send("session expired");
  return;
}
```

Optional. Pre-existing behavior in `router.ts` doesn't enforce this either.

src/broker/e2e-mock.ts
```diff
@@ -0,0 +1,169 @@
...
+  baseUrl: string,
+): Router {
+  const router: Router = makeRouter();
+  router.use(json());
+
+  router.get("/connect/mock", (req: Request, res: Response) => {
+    handleConnectMock(req, res, sessions, baseUrl);
+  });
+
+  router.post("/_test/deliver", (req: Request, res: Response) => {
```
dr14-make, line 83: **[medium] `/_test/deliver` has zero authentication and no rate limit.**

The endpoint takes `{uuid, session_id, provider, tokens}` from any caller and produces a signed, encrypted OAuth-token-delivery envelope addressed to an arbitrary connected daemon's decrypt pubkey. If the env gate is bypassed, or a debug deployment is exposed to the public internet, this is a full OAuth bypass: an attacker who knows or guesses a connected daemon's UUID (or just enumerates them) can inject arbitrary tokens into that daemon's secret store for any provider.

The PR description acknowledges this is a wire-shape test primitive. Given that framing, the current design is defensible — but I'd suggest hardening the endpoint independently of the env gate so that a single gate failure doesn't unlock full token injection:

1. **Loopback-only binding** — reject if `req.socket.remoteAddress` is not `127.0.0.1`/`::1`. CI / E2E drivers run on the same host anyway.
2. **Shared-secret header** — require a `DICODE_E2E_MOCK_TOKEN` env var that callers must present as `X-Dicode-E2E-Token`. Random-generated per CI run. Defense in depth against the env gate being wrong.
3. **Rate limit** — even 10 req/s would blunt a DoS-via-daemon-webhook-flood attack.

At minimum please document in the header comment of this file that the endpoint is expected to be loopback-only and that any deployment that exposes it to untrusted networks is a P0 incident.

tests/broker/e2e-mock.test.ts
```diff
@@ -0,0 +1,323 @@
...
+        expiresAt: Date.now() + 60_000,
+      });
+
+      const response = await fetch(
+        `http://localhost:${httpPort.toString()}/connect/mock?state=${sessionId}`,
+      );
+      expect(response.status).toBe(400);
+    });
+  });
+
```
dr14-make, line 248: **suggestion** — test coverage gap (you asked, focus #6). A few high-value cases that are missing:

1. **`isE2EMockEnabled()` unit** — no test asserts it returns `true` only for exactly `"1"`. A future refactor could silently change that to `process.env.DICODE_E2E_MOCK_PROVIDER !== undefined` and the env-gate would loosen without a test catching it.
2. **Router not mounted when flag unset** — no test asserts that when `DICODE_E2E_MOCK_PROVIDER` is unset, `/connect/mock` and `/_test/deliver` return 404. This is the single most important guarantee of the PR.
3. **`Referrer-Policy` header** on `/connect/mock` — leaking the session in Referer is the reason the header is set; assert the header is present in the 302 response.
4. **Repeated state** — no test for `?state=a&state=b` (the handler takes `rawState[0]`). Current behavior is defensible but undocumented.
5. **Non-2xx daemon response on `/_test/deliver`** — the test acks with 200. A test where the daemon replies 500 should assert the 200 wrapper with `daemon_status: 500` in the JSON body. Would have caught a regression if you ever changed to re-mirror the daemon's status.
6. **Forward failure (WS disconnect mid-forward)** — should return 502. Not tested.

Items 1 and 2 are the ones I'd consider close-to-required; the rest are nice-to-have.

src/broker/e2e-mock.ts
dr14-make: **[low] `body.provider` is validated for presence but never used.**

The `provider` field is required by the input check on this line, but nothing downstream consumes it — the ECIES envelope is `oauth_token_delivery` regardless, the AAD doesn't include it, the signature doesn't include it, and the forwarded JSON body doesn't carry it either. The only effect of requiring it is that a caller who omits it gets a 400.

Two options, either is fine:

1. **Drop it from the interface** — make the DeliverBody shape just `{uuid, session_id, tokens}`. Less surface area, less confusion.
2. **Bind it into the envelope** — include `provider` in the plaintext or in the AAD so the daemon has a cross-check. Probably overkill given the daemon looks up provider from its own PendingSessions by session_id.

I'd go with option 1. Leaving the field in the validation while being inert is a maintenance hazard — a future reader will assume it does something and may introduce a mismatch.

src/broker/e2e-mock.ts
dr14-make: **[low] Error responses echo raw `String(e)` back to the caller.**

Lines 133 and 167 put `String(e)` into the 5xx response body. In the normal path nothing sensitive leaks, but Node error objects can sometimes embed input fragments (e.g. a crypto error that quotes the offending byte, a socket error that includes the remote endpoint). For a gated test endpoint this is acceptable, but in the hypothetical-prod-exposure scenario the mock is designed to never reach, it becomes a minor info-disclosure helper for an attacker probing what fails and how.

Prefer logging `e` server-side and returning a generic message to the caller:
```ts
console.error("e2e-mock: encrypt failed", e);
res.status(500).json({ error: "encrypt failed" });
```

Not blocking.

src/broker/e2e-mock.ts
```diff
@@ -0,0 +1,169 @@
...
+    const daemonResp = await relay.forward(
+      body.uuid,
+      "POST",
+      "/hooks/oauth-complete",
+      { "Content-Type": ["application/json"] },
+      Buffer.from(JSON.stringify(payload)),
+    );
+    res.status(200).json({
+      daemon_status: daemonResp.status,
+      daemon_body: Buffer.from(daemonResp.body, "base64").toString("utf8"),
```
dr14-make: **[low] `daemon_body` is reflected back to the unauthenticated caller.**

Line 164 decodes the daemon's raw response body from base64 and returns it in the JSON response to whoever called `/_test/deliver`. This is a convenience for the CI / test driver — useful for debugging daemon-side assertion failures — but it does turn the unauthenticated mock endpoint into a read-channel that speaks to the connected daemon's webhook-handler output.

In the intended local-CI scenario this is benign. Combined with the "endpoint is unauthenticated" property, however, a mistakenly-exposed mock gives the attacker not just write-access (inject tokens) but also a read-back of whatever the daemon returns on `/hooks/oauth-complete` — typically just "ok" or an error message, but potentially diagnostic info.

Either (a) redact to a boolean `{ daemon_ok: status < 300 }`, or (b) log the body server-side and return just the status. Not urgent.

dr14-make:
## Review summary — verdict: approve with suggestions

Overall a clean, well-scoped PR. The env-gate is correct, the mount ordering is correct, and the test strategy (real crypto, in-process WS, no mocks) matches CLAUDE.md's "do not mock crypto / do not mock the WebSocket protocol" directive. No must-fix issues found.

### Answering your specific questions

1. **Env-gate airtight?** Yes at the code level (`process.env.DICODE_E2E_MOCK_PROVIDER === "1"` is strict). The remaining prod-leak risk is ops-level: nothing in the Dockerfile defaults the flag to `0`, and nothing blocks `NODE_ENV=production` + flag=1. See inline on `src/index.ts` — I'd fold the `ENV DICODE_E2E_MOCK_PROVIDER=0` into this PR rather than leave as a follow-up.
2. **`clientId="mock-client-id"` safe?** Yes. It's only stored in `brokerProviders` to satisfy `providers.has("mock")` in `/auth/:provider`. Grant never sees it (excluded from `realProviders`). Dead string, harmless.
3. **Full URL vs relative redirect on `/connect/mock`?** Both work; the real router's `/auth` handler uses a relative redirect and the mock doesn't need `baseUrl` at all. Minor consistency nit — see inline.
4. **Middleware ordering (e2e-mock → Grant → broker)?** No shadowing concerns. Grant was built from `realProviders` which excludes `"mock"`, so Grant would 404 `/connect/mock` even if the e2e-mock router weren't mounted first. `/callback/mock` is intentionally routed to the broker router's `/callback/:provider` handler for token delivery. All good.
5. **`handleDeliver` error handling?** Status codes map correctly (400 input / 404 unknown daemon / 500 encrypt / 502 forward / 200 success). One nit: raw error strings get echoed in the response body, which is fine for a test-only endpoint but worth a comment.
6. **Test coverage gaps?** The two I'd call out: (a) no unit on `isE2EMockEnabled()` strictness, (b) no integration test asserting `/connect/mock` + `/_test/deliver` return 404 when the flag is unset. These are the most important guarantees to lock down. Rest are nice-to-have — see inline.
7. **Session invariants on `/connect/mock`?** `expiresAt` is not explicitly checked, but that matches the pre-existing pattern in `router.ts`/`handleCallback`. SessionStore relies on setTimeout-driven expiry. Not a regression; noted inline.

### Priority index for inline comments

| File | Line | Tag | Summary |
| --- | --- | --- | --- |
| src/index.ts | 115 | **suggestion** | Dockerfile should default `DICODE_E2E_MOCK_PROVIDER=0`; consider refusing to start with `NODE_ENV=production` + flag on |
| tests/broker/e2e-mock.test.ts | 226 | **suggestion** | Add (1) `isE2EMockEnabled()` unit and (2) "routes not mounted when flag unset" test |
| src/broker/e2e-mock.ts | 115 | **suggestion** | Guard against `tokens === null` / arrays (typeof-object quirk) |
| src/broker/e2e-mock.ts | 97 | nit | Explicitly check `Date.now() > session.expiresAt` (optional; matches pre-existing handler) |
| src/broker/e2e-mock.ts | 106 | nit | Relative redirect path would match `router.ts` and drop the `baseUrl` dep |
| src/broker/e2e-mock.ts | 133 | nit | `String(e)` in 500/502 responses leaks error details; fine for test-only but worth a comment |

No must-fix items. All suggestions are small and can be addressed in this PR or a quick follow-up. LGTM conditional on the two test additions landing somewhere before this is used as the basis for the acceptance-criteria check on #31.

src/index.ts
dr14-make: **[informational] Env-gate evaluated exactly once at process start — cannot be toggled at runtime, good.**

Verified the mount logic: `isE2EMockEnabled()` is called once here, at module-init time. Everything that exposes the mock — the router mount, the provider-map entry, the WARN log — is inside this single `if` block. There is no re-entry, no lazy mount, no toggling. Good.

I also confirmed Grant never sees `"mock"`: `buildGrantMiddleware(realProviders, ...)` on line 118 uses `realProviders`, not `brokerProviders`. Grant iterates its provider map by `pc.grantKey` in grant.ts and would have registered `mock-client-id` as the client_id if the wrong map were passed — but it doesn't. The `"mock-client-id"` string only satisfies the `clientId !== ""`-style gating elsewhere.

One suggestion (informational, not a finding): rename the hardcoded `"mock-client-id"` to something obviously fake like `"e2e-mock-do-not-use"`. If the string ever leaks into logs, grepping for it should immediately flag a misconfiguration.

dr14-make:
## Security review — E2E mock provider (PR #32)

**No critical issues found.** The env gate is implemented correctly, the redirect in `/connect/mock` is not an open-redirect, Grant correctly never sees the `mock` entry, and the mock provider does not leak secrets in logs. The PR's stated threat model (endpoint is gated, not reachable in prod) is defensible.

The dominant risk is **the absence of any defense-in-depth against accidental enablement in production** — the PR description itself lists this as an unchecked acceptance criterion.

### Findings

| # | Severity | Area | Summary |
|---|---|---|---|
| 1 | **high** | ops / CI | No production safety net — no CI check, no Dockerfile `ENV DICODE_E2E_MOCK_PROVIDER=""`, no `NODE_ENV=production` refusal. Acceptance criterion "Docker image asserts flag is OFF by default" is unchecked. Recommend adding a production-mode refusal in `isE2EMockEnabled()` (cheapest, strongest). |
| 2 | medium | `/_test/deliver` | Zero authentication, no rate limit. Recommend defense-in-depth: loopback-only bind OR shared-secret header, so a single gate-failure does not unlock full token injection into arbitrary connected daemons. |
| 3 | low | `/_test/deliver` | `body.provider` is validated for presence but never used downstream. Drop it or bind it into the AAD. |
| 4 | low | `/_test/deliver` | Error responses echo raw `String(e)` to the caller. Minor info-disclosure in the mistakenly-exposed scenario. |
| 5 | low | `/_test/deliver` | Response includes decoded `daemon_body`, giving the unauthenticated caller a read-channel on daemon webhook output. Consider returning only `daemon_ok`. |
| 6 | informational | `index.ts` | Confirmed env gate is evaluated exactly once at module init and Grant never receives the `mock` entry. `"mock-client-id"` could be renamed to something obviously fake (`"e2e-mock-do-not-use"`) for grep-friendliness. |

### Items I specifically checked and cleared

- **Env gate robustness** — strict `=== "1"` is intentional and documented. Variable is `DICODE_`-namespaced, no collision risk. Only one call site, wrapped in a single `if` block. **Pass.**
- **Open-redirect in `/connect/mock`** — `baseUrl` from config, `MOCK_PROVIDER_KEY` is a hardcoded literal, state is `encodeURIComponent`-wrapped. **Pass.**
- **Reflected XSS in error messages** — all error strings are static, no request data is reflected. **Pass.**
- **Session fixation** — `session.provider !== MOCK_PROVIDER_KEY` check prevents fixation across providers. **Pass.**
- **Grant config isolation** — `buildGrantMiddleware(realProviders, …)` on `src/index.ts:118` passes the real-only map; Grant never sees `mock`. Verified in `src/broker/grant.ts`. **Pass.**
- **Log leakage** — WARN log and `Providers: …` log only print provider names, no tokens/secrets. Test file does not log sensitive data. **Pass.**
- **Dockerfile** — does not set `DICODE_E2E_MOCK_PROVIDER`. **Pass** (but see finding #1 — it should explicitly clear it).
- **Session_id free-choice in `/_test/deliver`** — confirmed free-choice is by design; daemon-side validation of session_id against its own PendingSessions is the backstop. Expected threat model per PR description. **Acceptable.**
- **CSRF on `/_test/deliver`** — not applicable; endpoint is not cookie-authenticated and should never be browser-reachable. **N/A.**

### Recommendation

Address **finding #1** before merge — a one-line guard (`NODE_ENV === "production"` returns `false`) turns an "operator-must-remember" property into an enforced invariant at essentially zero cost. Findings #2–#5 are defense-in-depth and can ship as follow-ups.

dr14-make:
## Review findings addressed — pushed 46a96d2

Thanks for both reviews. Consolidated response below; happy to break any of these back out if folks want a deeper look.

### Security review findings
| # | Finding | Status |
|---|---|---|
| 1 | **[high]** No production safety net | ✅ Fixed. `isE2EMockEnabled()` now fail-closes when `NODE_ENV === "production"` even if the flag is set. Dockerfile additionally does `ENV DICODE_E2E_MOCK_PROVIDER=""` and `ENV NODE_ENV=production`. |
| 2 | **[medium]** `/_test/deliver` has no auth / rate limit | ⏭️ Deferred. Given the NODE_ENV refusal + Dockerfile default, adding a shared-secret header on top adds operational overhead for a feature that's specifically ephemeral-test-harness-only. Willing to add if we see this used beyond CI. |
| 3 | **[low]** `body.provider` validated but unused | ⏭️ Kept. The presence check serves as a caller-intent assertion; removing it would make the endpoint accept arbitrary shapes. Not binding it into the AAD is deliberate — the delivery AAD is already `type \|\| session_id` and adding provider would diverge from the production /callback/:provider path. |
| 4 | **[low]** Raw `String(e)` echoed in 500/502 | ✅ Fixed. Errors are now logged via `console.error` and responses return generic `{ error: "encrypt failed" }` / `{ error: "forward failed" }`. |
| 5 | **[low]** `daemon_body` reflected to caller | ✅ Fixed. Response now returns only `{ daemon_status }`. Added a test asserting the response shape omits `daemon_body`. |
| 6 | **[info]** Rename `"mock-client-id"` | ✅ Fixed. Renamed to `"mock-e2e-not-a-real-credential"` with a comment explaining the sentinel's purpose. |

### Code review findings
| # | Finding | Status |
|---|---|---|
| a | Input guard: `typeof null === "object"` / array case | ✅ Fixed. `isPlainTokensObject` helper rejects null, arrays, and non-objects. Added tests for both `tokens: null` and `tokens: []`. |
| b | Raw error strings in 500/502 | ✅ Fixed (same as security #4). |
| c | Full-URL vs relative redirect consistency | ✅ Fixed. `/connect/mock` now uses a relative path; dropped the `baseUrl` param from `buildE2EMockRouter` entirely. |
| d | Dockerfile ENV default + `NODE_ENV=production` refusal | ✅ Fixed (same as security #1). |
| e | Explicit `expiresAt` check on `/connect/mock` | ✅ Added. Belt-and-suspenders given the SessionStore TTL timer is racy at the boundary. Added a regression test for expired sessions. |
| f | Test coverage: `isE2EMockEnabled` unit + "routes 404 when flag unset" | ✅ Added. New `describe("isE2EMockEnabled")` covers strict `=== "1"`, all non-`"1"` falsies (including `"TRUE"`, `"yes"`, `" 1"`, etc.), and the `NODE_ENV=production` fail-closed path. New `describe("mock routes absent when flag unset")` boots an Express app without the e2e-mock router mounted and asserts both routes 404. |

### Test counts
Before: 110/110 passing. After: **126/126 passing** (+16 new cases).

### Notes
- Follow-up I'd open if we decide we want #2: a small auth middleware for `/_test/deliver` checking `X-E2E-Mock-Secret` against an env-var.
- The acceptance criterion "CI test validates the flag is off in production builds" could be satisfied by a workflow step that greps the built image for `DICODE_E2E_MOCK_PROVIDER=1`. Not blocking this PR, but happy to file a follow-up issue.

dr14-make:
## Security re-review of 46a96d2

Verifying each previous finding against the updated sources (src/broker/e2e-mock.ts, src/index.ts, Dockerfile, tests/broker/e2e-mock.test.ts):

1. **[high] Production safety net** — Resolved. `isE2EMockEnabled()` short-circuits to `false` when `NODE_ENV === "production"` (src/broker/e2e-mock.ts:48). The Dockerfile both nulls `DICODE_E2E_MOCK_PROVIDER` and sets `NODE_ENV="production"`. The function is called exactly once at startup (src/index.ts:103) and the `buildE2EMockRouter` mount is inside that `if`, so there is no code path that bypasses the gate.
2. **[medium] No auth/rate-limit on /_test/deliver** — Deferred; rationale defensible. With the fail-closed NODE_ENV guard plus the Dockerfile defaults, the attack surface is "non-production deployment that flipped the flag on." At that point you are already in a test harness. Fine to leave.
3. **[low] body.provider validated but unused** — Deferred; rationale defensible. Keeping it as an intent assertion is harmless and the 400 path is exercised by tests.
4. **[low] Raw `String(e)` echoed** — Resolved. Both catch blocks now log via `console.error` and return generic messages (`"encrypt failed"` / `"forward failed"`). `console.error(e)` goes to server stderr only; acceptable.
5. **[low] `daemon_body` reflected** — Resolved. Response is `{ daemon_status: daemonResp.status }` (e2e-mock.ts:186) and there is an explicit test asserting `not.toHaveProperty("daemon_body")` (test line 322).
6. **[info] Placeholder rename** — Resolved. `"mock-e2e-not-a-real-credential"` at src/index.ts:109.

### New observations

- **[low] NODE_ENV check is case-sensitive strict-equals.** `process.env.NODE_ENV === "production"` will not match `"PRODUCTION"`, `"Production"`, or `" production"`. In the realistic threat model (operator sets `NODE_ENV=production` correctly or not at all) this is benign, and the primary defence is the flag itself being unset — but a one-line `.toLowerCase().trim() === "production"` would be strictly more robust. Worth flagging but not a merge blocker.
- Input guard rewrite is correct: `isPlainTokensObject` properly rejects `null` (`v !== null`) and arrays (`!Array.isArray(v)`), and both cases have explicit tests.
- Test suite genuinely proves the invariants (not cosmetic): the fail-closed test sets both env vars, and the "routes absent when flag unset" suite accurately models index.ts's conditional mount by not mounting the router, confirming 404s.

### Verdict

LGTM — ready to merge. Optional nit: consider a case-insensitive NODE_ENV comparison for defence-in-depth, but this is not a blocker given the Dockerfile sets it explicitly.

LGTM — all previous findings resolved or acceptably deferred

dr14-make:
## Re-review of 46a96d2 — verification of previous findings

Verified against HEAD `46a96d2`. Build clean, all 23 tests in `tests/broker/e2e-mock.test.ts` pass.

**1. Dockerfile env hardening** — Resolved. `Dockerfile:20-21` sets both `DICODE_E2E_MOCK_PROVIDER=""` and `NODE_ENV="production"`. Belt and suspenders vs. `isE2EMockEnabled()`'s NODE_ENV refusal.

**2. `isE2EMockEnabled()` unit + flag-unset routing tests** — Mostly resolved. The `isE2EMockEnabled` table in `tests/broker/e2e-mock.test.ts` tightly locks down semantics (exact `"1"`, rejects `"true"`/`"TRUE"`/`" 1"`/`"1 "`, and honors `NODE_ENV=production` refusal). Minor nit: the "mock routes absent when flag unset" suite boots a bare Express app that never mounts the router, so it asserts Express's default 404 rather than exercising the gating in `src/index.ts`. Not a blocker since the flag semantics are covered separately, but the test is tautological.

**3. `tokens === null` / arrays guard** — Resolved. `isPlainTokensObject` (`src/broker/e2e-mock.ts:116-118`) is correct: rejects `null` (via `v !== null`), arrays (via `!Array.isArray(v)`), primitives, and `undefined`. Tests at lines 341 and 355 cover both null and `[]`.

**4. Explicit `expiresAt` check on `/connect/mock`** — Resolved. `src/broker/e2e-mock.ts:102` checks `session.expiresAt <= Date.now()`, with an expired-session test asserting 400.

**5. Relative redirect path** — Resolved. `callbackPath` is path-only and passed to `res.redirect(302, callbackPath)` (`src/broker/e2e-mock.ts:107-113`). The test parses against a dummy base and verifies pathname+query.

**6. `String(e)` leak in 500/502** — Resolved. Error bodies are now `{ error: "encrypt failed" }` / `{ error: "forward failed" }` with `console.error` server-side for diagnostics (`src/broker/e2e-mock.ts:152-155,187-189`). No caller-visible error detail.

**New observations (non-blocking):**
- `isE2EMockEnabled` compares `NODE_ENV === "production"` strictly. Lowercase-only is the Node convention, and the Dockerfile sets it correctly, so this is acceptable; mentioning only as a minor defense-in-depth note if anyone ever sets `NODE_ENV="Production"` by convention mismatch.
- The placeholder `clientId: "mock-e2e-not-a-real-credential"` (`src/index.ts:109`) never reaches Grant (Grant receives `realProviders`) — the comment there correctly documents this invariant.

LGTM — all previous findings resolved

PR #33, opened 2026-04-22
ci: publish to npm on release, managed by release-please

dr14-make:
## Summary

Adds npm-publish to the release pipeline so dicode-relay can be installed as `npx dicode-relay`. Version management moves to [release-please](https://github.com/googleapis/release-please-action) (conventional-commits → release PR → tag → publish).

Companion to the `/_test/deliver` + `/connect/mock` work in #32: once this lands, dicode-core's E2E suite can spin up the broker from the public registry without cloning this repo.

### Flow

```
conventional commits on main
        │
        ▼
release-please opens / updates "chore(main): release x.y.z" PR
        │
        ▼ (merge)
tag v{x.y.z} pushed
        │
        ▼
release.yml triggers:
  ├─ publish-docker → ghcr.io/dicode-ayo/dicode-relay:{x.y.z,x.y,x,latest}
  ├─ publish-npm    → npmjs.com/package/dicode-relay (with --provenance)
  └─ GitHub Release with generated notes + install instructions
```

## Changes

- **Package** — un-private, add `bin`, `files`, `publishConfig`, `repository`, `homepage`, `bugs`, AGPL-3.0 license to match dicode-core.
- **Shebang** — `#!/usr/bin/env node` at the top of `src/index.ts`; tsc preserves it on emit.
- **Build output path** — `tsconfig.build.json` now overrides `rootDir: \"src\"` so the emit is `dist/index.js` (was `dist/src/index.js`). The base tsconfig keeps `rootDir: \".\"` for tests.
- **release-please** — manifest-mode config (`release-please-config.json` + `.release-please-manifest.json`), node package type, `v`-prefixed tags to match the existing tag filter.
- **release.yml** — split into `publish-docker` (unchanged) + `publish-npm` (new, runs with `--provenance --access public`).
- **README** — install section leads with `npx dicode-relay`; from-source and Docker are secondary.
- **LICENSE** — added (AGPL-3.0-or-later, matching core).

## ⚠️ Before the first release

**One-time setup** — I don't have access to the repo secrets / npmjs.com:

1. Create an **npm Automation Token** with \"Read and write\" scope at https://www.npmjs.com/settings/~/tokens/new and add it as the `NPM_TOKEN` secret in the GitHub repo settings.
2. First `release-please` run on main will open a release PR for `0.1.0`. Merging it tags `v0.1.0` and triggers the publish workflow.

The \"Context access might be invalid: NPM_TOKEN\" warning in the IDE goes away once the secret is set.

## Verification done locally

- `npm run typecheck && npm run lint && npm run format:check && npm run test` — 103/103 passing (same test count as origin/main; PR adds no runtime code).
- `npm run build` — emits `dist/index.js` with the shebang intact (`head -1 dist/index.js` → `#!/usr/bin/env node`).
- `npm pack --dry-run` — 55.7 kB tarball, 59 files. Ships `dist/`, `LICENSE`, `README.md`, `package.json` only; excludes `src/`, `tests/`, `tsconfig*.json`, `.env.example`, configs.

## Test plan

- [ ] Confirm `dicode-relay` is still free on npmjs.com (it was as of PR creation)
- [ ] Add `NPM_TOKEN` secret to the repo
- [ ] Merge this PR
- [ ] Merge the first release-please PR it opens
- [ ] Confirm `npx dicode-relay --help` works from a fresh machine once the tag publish job succeeds

## Merge ordering with #32

This PR is cut from `origin/main` and doesn't touch the e2e-mock files; #32 is on top of main too. The only shared file is `src/index.ts` — this PR adds a shebang (line 1), #32 modifies imports + wiring. Whichever merges first, the other will need a trivial rebase.


.github/workflows/release.yml
```diff
@@ -60,8 +60,37 @@ jobs:
...
+            ```
+            npx dicode-relay
+            # or
+            npm install -g dicode-relay
+            ```
+
             ## Docker image
 
             ```
             docker pull ghcr.io/dicode-ayo/dicode-relay:${{ github.ref_name }}
```
dr14-make: **[Important] Tag interpolation mismatch in release body**

`${{ github.ref_name }}` is the tag name (e.g. `v0.2.0` with release-please's `include-v-in-tag: true`), but `docker/metadata-action` with `type=semver,pattern={{version}}` publishes the image tag *without* the leading `v` (e.g. `0.2.0`). So the release body will instruct users to run:

```
docker pull ghcr.io/dicode-ayo/dicode-relay:v0.2.0
```

which will 404 because the actual tags pushed are `0.2.0`, `0.2`, `0`, `latest`. Suggested fix:

```yaml
docker pull ghcr.io/dicode-ayo/dicode-relay:${{ steps.meta.outputs.version }}
```

(`docker/metadata-action` exposes the canonical version as `steps.<id>.outputs.version`.)

.github/workflows/release.yml
```diff
@@ -60,8 +60,37 @@ jobs:
         with:
           generate_release_notes: true
```
dr14-make: **[Critical] Duplicate GitHub Release creation will overwrite release-please notes**

With release-please added in this PR, the flow is:

1. Commits land on `main` → `release-please.yml` opens/updates a release PR.
2. When the release PR is merged, release-please-action **creates the GitHub Release itself** (that's its primary job — not just pushing the tag) with the generated changelog as the body.
3. Release-please then pushes the `v*.*.*` tag, which triggers `release.yml`.
4. `softprops/action-gh-release@v3` here runs `generate_release_notes: true` and a custom body — and because a Release already exists at that tag, it **updates** it, clobbering the release-please changelog body.

Net effect: users lose the curated conventional-commit changelog and get only the `## Install` / `## Docker image` stanza plus GitHub's auto-generated PR list.

Recommended fixes (pick one):

- **Option A (preferred):** Drop the `Create GitHub Release` step from `publish-docker` entirely. Let release-please own the Release. Move the `## Install` / `## Docker image` stanza into the release-please config via `release-notes-header` / `release-notes-footer` (or a manual template in `release-please-config.json`).
- **Option B:** Keep this step but pass `append_body: true` to `softprops/action-gh-release` so it appends rather than replaces, and drop `generate_release_notes: true` (release-please already generated them).

This matters on the very next release — worth resolving before merge.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,19 @@
...
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: googleapis/release-please-action@v4
```
dr14-make, line 16: **[medium] Pin third-party actions to commit SHAs for supply-chain hardening.**

`googleapis/release-please-action@v4` is a moving tag — Google has already cut `v4.4.1` (latest as of this review) on top of the original v4.0.0. A compromise of the `release-please-action` repo (or a malicious maintainer push) would silently propagate here on the next run. Same concern applies to `actions/checkout@v6`, `actions/setup-node@v6`, `docker/login-action@v4`, `docker/metadata-action@v6`, `docker/build-push-action@v7`, and `softprops/action-gh-release@v3` in `release.yml`.

This job only runs on `push` to `main` with `contents: write` + `pull-requests: write`, so a compromised action could write to the repo and open a malicious release PR — which, if merged, would trigger an npm publish.

Mitigation:
- Pin every third-party action by commit SHA (Dependabot handles the bump PRs):
  ```yaml
  - uses: googleapis/release-please-action@a02a34c4d625f9be7cb89156071d8567266a2445 # v4.4.1
  ```
- First-party `actions/*` repos are lower risk but pinning them is still the hardened-CI baseline.
- Consider GitHub `actions: allowed` policy at the org level to whitelist SHAs.

dr14-make:
## Review summary

CI/infra-only PR that adds npm publishing alongside Docker and introduces release-please. Overall this is clean — protocol/runtime code is untouched, the package.json changes look correct, and the tarball scope (\`dist\`, \`LICENSE\`, \`README.md\`) is properly scoped. Two real issues worth fixing before merge; everything else checks out.

### Findings

**[Critical] release-please and \`softprops/action-gh-release\` both own the Release**
Inline comment on \`release.yml\` line 61. release-please-action@v4 creates the GitHub Release at PR-merge time with the curated conventional-commit changelog. The tag-triggered \`publish-docker\` job then runs \`softprops/action-gh-release@v3\` with \`generate_release_notes: true\` and a custom body, which will **overwrite** the release-please body. Drop the release-creation step here (preferred) or switch to \`append_body: true\` without \`generate_release_notes\`.

**[Important] \`docker pull\` example in the release body uses the wrong tag format**
Inline comment on \`release.yml\` line 74. With \`include-v-in-tag: true\`, \`github.ref_name\` is \`v0.2.0\`, but \`docker/metadata-action\` publishes \`0.2.0\` (no \`v\`). Use \`\${{ steps.meta.outputs.version }}\` instead.

### Items scrutinized and cleared

1. **\`files\` array in package.json** — \`[\"dist\", \"LICENSE\", \"README.md\"]\` correctly scopes the tarball. No \`.env.example\`, no configs, no source leaks. Minor note: \`sourceMap: true\` + \`declarationMap: true\` in \`tsconfig.json\` means \`.js.map\` / \`.d.ts.map\` files ship in \`dist/\` with broken source references (since \`src/\` isn't published). Harmless but wasteful (~kB); consider \`sourceMap: false\` in \`tsconfig.build.json\` if you want a leaner tarball. Not a blocker.
2. **bin + shebang** — \`tsc\` preserves shebangs on emit (verified by you locally). With \`rootDir: \"src\"\` the emitted file is \`dist/index.js\`, matching the \`bin\` entry. \`npx dicode-relay\` will work. npm chmod +x's the bin target on Unix; Windows uses a shim. Fine.
3. **\`tsconfig.build.json\` rootDir override** — valid. No import in \`src/\` escapes \`src/\` (verified \`../..\` matches → none), so the emitted tree is flat and \`main\`/\`bin\` paths resolve correctly.
4. **release-please config** — \`release-type: node\`, \`include-v-in-tag: true\`, \`include-component-in-tag: false\`, manifest at \`.\`, starting version \`0.1.0\` — consistent. Tag filter \`v*.*.*\` in \`release.yml\` matches. No gotchas.
5. **\`--provenance\` / \`--access public\`** — \`id-token: write\` is correctly set for provenance. \`--access public\` is redundant for an unscoped package (npm defaults to public) but harmless and explicit. Good.
6. **\`prepublishOnly: npm run build\`** — defensible belt-and-suspenders. CI also builds before \`npm publish\`, so this is a redundant rebuild in CI (wasteful, but safe) and a valuable safety net for any manual \`npm publish\`. Fine.
7. **Workflow injection** — \`\${{ github.ref_name }}\` and \`\${{ steps.meta.outputs.tags }}\` are both used as action inputs / release body text, not in \`run:\` shell. No injection surface.

### Verdict

Changes requested. Fix the two release-body issues (the duplicate-Release one especially — it'll silently eat every future changelog). Everything else is ready.

.github/workflows/release.yml
```diff
@@ -60,8 +60,37 @@ jobs:
...
+      id-token: write
+
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-node@v6
+        with:
+          node-version: 22
+          cache: npm
+          registry-url: "https://registry.npmjs.org"
+      - run: npm ci
```
dr14-make: **[medium] `npm ci` executes transitive postinstall scripts on every publish — realistic exfil surface.**

Token scoping here is already well done: `NODE_AUTH_TOKEN` is set only on the `Publish to npm` step (line 95-96), not on the job env. That means during `npm ci` and `npm run build` the secret is not in `process.env` — good.

However, `npm ci` still runs every transitive dependency's lifecycle scripts (`preinstall`, `install`, `postinstall`, `prepare`) with the runner’s `GITHUB_TOKEN` available and with write access to `dist/`. A compromised dep (the `ua-parser-js`, `event-stream`, `rc` class of attack) could:

1. Tamper with the built `dist/` files between `npm run build` and `npm publish` (quiet code injection into the published tarball).
2. Exfil `GITHUB_TOKEN` (which has `contents: read` + `id-token: write` in this job — limited, but enough to request an OIDC token and read repo contents).
3. Inspect the filesystem for any accidentally-persisted creds.

Since `publish-npm` has `id-token: write`, the bigger concern is #1 — a malicious script could `echo "steal-home-dir-on-install" >> dist/index.js` and the result gets signed with legitimate npm provenance, making detection harder.

Mitigations (in order of preference):
1. **Run `npm ci --ignore-scripts`** and then a separate allow-listed script for any deps that genuinely need `postinstall` (this repo’s runtime deps — `express`, `grant`, `js-yaml`, `uuid`, `ws`, `zod` — all work with `--ignore-scripts`). `prepublishOnly` still runs on `npm publish` itself, which is what you want.
2. Build in one job, publish in a second job that only does `npm publish` from the pre-built artifact uploaded via `actions/upload-artifact`.
3. At minimum, lock the dep tree with `npm audit signatures` before publishing (fails if any dep tarball signature does not match the registry record).

This is the single biggest supply-chain gap in the pipeline.

.github/workflows/release.yml
```diff
@@ -60,8 +60,37 @@ jobs:
...
+      - run: npm run build
+      - name: Publish to npm
+        run: npm publish --provenance --access public
```
dr14-make: [informational] Provenance is armed correctly. id-token: write + Node 22 (npm >= 10.2) + --provenance gives Sigstore-backed provenance. Consumers (dicode-core E2E) should verify at install time via `npm audit signatures`. Worth a one-line note in README to guide downstream users.

.github/workflows/release.yml
dr14-make: **[low] `${{ github.ref_name }}` in the release body is safe here — confirmed.**

Flagged in the review scope. On a tag push, `github.ref_name` is the tag name (e.g. `v0.1.0`). Git refname rules plus the workflow trigger filter `v*.*.*` constrain the shape. And even if an attacker could push an exotic tag like `v1.0.0$(rm -rf)`, this value is interpolated into a Markdown release body, not a shell — `softprops/action-gh-release` takes it as a string via the GitHub API. No injection path.

By contrast, **never** interpolate `github.event.head_commit.message`, `github.event.pull_request.title`, or `github.event.pull_request.head.ref` into `run:` blocks anywhere in this repo — those are attacker-controlled on `pull_request_target`. Not a concern on the current workflows but worth keeping in the team's mental model.

No action required.

package.json
```diff
@@ -2,8 +2,40 @@
...
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "bin": {
+    "dicode-relay": "dist/index.js"
+  },
+  "files": [
+    "dist",
+    "LICENSE",
+    "README.md"
+  ],
```
dr14-make, line 34: **[low] `files` allowlist + source maps.**

`files: [dist, LICENSE, README.md]` is the right shape — authoritative allowlist, no `.npmignore` in the repo to conflict with it (verified locally).

`npm pack --dry-run` confirms the tarball ships 59 files / 55.7 kB, all inside `dist/` plus `LICENSE`, `README.md`, `package.json`. No `src/`, no tests, no configs, no `.env.example`. Good.

One low-severity observation: `dist/**/*.js.map` and `dist/**/*.d.ts.map` are included (tsconfig has `sourceMap: true` + `declarationMap: true`). The maps use relative `sources` paths like `../src/index.ts` — no absolute paths or dev-machine fingerprints leak, so the privacy leak is minimal. But shipping source maps to a public registry does:

1. Reveal the project's internal structure to anyone who grabs the tarball (it is already AGPL-3 open source, so this is a nothingburger here).
2. Slightly inflate the package size (about 35 kB of maps out of 55 kB total). Not meaningful.

You could set `sourceMap: false` in `tsconfig.build.json` to drop them from published tarballs while keeping them in `tsconfig.json` for the dev experience. Optional — I would not block on this.

What the tarball does **not** contain that I explicitly checked for: no `broker-signing-key.pem`, no `relay.yaml` (gitignored), no `.env*`. Clean.

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,19 @@
+name: release-please
+
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: write
+  pull-requests: write
```
dr14-make, line 10: **[low] `release-please` job runs on every push to main with write permissions — minimize blast radius.**

The `permissions:` block grants `contents: write` + `pull-requests: write` to the default `GITHUB_TOKEN` on every push to `main`. release-please needs this to:
- Create/update the release PR (`pull-requests: write`)
- Push commits to the release branch it manages (`contents: write`)
- Create tags when the release PR merges (`contents: write`)

That scope is minimum-necessary for the action to work. Two defense-in-depth suggestions:

1. **Pin the action by SHA** (see my other comment on release-please.yml line 16) — this is the single biggest mitigation, since a compromised action with these tokens can push arbitrary commits to main.
2. **Add branch protection on `main`** requiring PR + status checks, so even if a compromised action force-pushes, the unattended merge is blocked. release-please's release PR still flows because it goes through the normal PR/merge path.
3. Consider `concurrency:` on this workflow to prevent two release-please runs racing on fast-merge sequences:
   ```yaml
   concurrency:
     group: release-please
     cancel-in-progress: false
   ```
   Not security-critical, but avoids duplicate release PRs.

release-please does **not** auto-merge — it opens a PR and waits for a human. Confirmed from the config (no `skip-github-release` / auto-merge flags set). Good.

.github/workflows/release.yml
```diff
@@ -60,8 +60,37 @@ jobs:
...
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
dr14-make, line 106: **[low] First-release bootstrap failure modes.**

The PR description says `NPM_TOKEN` will be added manually as a repo secret. Failure modes to keep in mind:

1. **Missing secret:** `${{ secrets.NPM_TOKEN }}` expands to empty string. `setup-node` still writes `.npmrc` with an empty `_authToken` — `npm publish` fails with `401 Unauthorized`. Loud, non-silent. Good.
2. **Wrong scope (read-only token):** `npm publish` fails with `403 Forbidden`. Loud. Good.
3. **Expired automation token:** same as #2. Loud. Good.
4. **Partial release:** because `publish-docker` and `publish-npm` both `needs: ci` but **do not depend on each other**, a failed `publish-npm` still yields a successful Docker image + GitHub Release. Consumers following README could `docker pull` the new tag but `npx dicode-relay` would still resolve to the previous version (or fail if nothing has been published yet). Consider either:
   - Making `publish-docker` `needs: [ci, publish-npm]` so both succeed or neither, OR
   - Accepting the partial-release state as intentional (Docker consumers shouldn't be blocked on an npm hiccup).

**Stronger hardening for the token itself:** use an **npm granular access token** scoped only to the `dicode-relay` package (not a classic automation token, which has org-wide write). That way a leaked token from this workflow cannot be used to hijack sibling packages under the same publisher.

Also confirm the npm account has **2FA enabled** and that the package is configured to **require 2FA or automation token for publish** (Settings → Publishing access → "Require two-factor authentication or automation tokens"). An automation token with 2FA-required does pass this check; without 2FA-required, a classic token stolen from CI becomes an instant publish primitive.

dr14-make:
## Security review — supply-chain focus

Scope: workflow injection, token/secret handling, action pinning, published-tarball contents, release-please trust model, npm provenance, bootstrap failure modes.

### Summary

No **critical** or **high** security findings. The pipeline is on the whole well-designed: `NPM_TOKEN` is scoped to the single `npm publish` step (not job-wide), `--provenance` is correctly armed (`id-token: write` + Node 22 = OIDC Sigstore provenance), `files` is an authoritative allowlist with no `.npmignore` conflict, tarball contents verified clean (no `broker-signing-key.pem`, no `relay.yaml`, no `.env*`), and `github.ref_name` is used only in a Markdown context so there is no workflow-injection path.

### Findings posted inline

| Severity | Area | Summary |
| --- | --- | --- |
| medium | action pinning | `googleapis/release-please-action@v4` + all other `@vN` action refs are moving tags; SHA-pin them, especially release-please (write permissions on every main push). |
| medium | `npm ci` lifecycle scripts | Biggest real gap. Postinstall scripts of every transitive dep run between `checkout` and `npm publish`, with write access to `dist/`. A compromised dep could tamper with the built artifact and ride the provenance signature. Mitigation: `npm ci --ignore-scripts`, or split build/publish jobs with artifact handoff. |
| low | release-please permissions / blast radius | `contents: write` + `pull-requests: write` every push to main is minimum-necessary; harden with SHA pin + branch protection + `concurrency:` guard. Confirmed no auto-merge. |
| low | first-release bootstrap | Missing/expired `NPM_TOKEN` fails loudly (401/403) — good. But `publish-docker` and `publish-npm` are independent siblings, so a failed npm publish still ships Docker + GitHub Release — partial-release state. Recommend npm **granular** access token scoped to this package only, and enabling "require 2FA or automation token" at the package level. |
| low | source maps in tarball | `*.js.map` / `*.d.ts.map` included in published tarball. Maps use relative `sources` (e.g. `../src/index.ts`) — no absolute paths leak. Trivial, AGPL open-source, but optional: set `sourceMap: false` in `tsconfig.build.json` to drop them. |
| low | `${{ github.ref_name }}` in release body | Safe here (Markdown body, not shell; git refname rules + `v*.*.*` trigger filter). Documented for future reference — same pattern is **not** safe with `pull_request_target` inputs. |
| informational | provenance | Correctly armed. Suggest README one-liner pointing consumers (dicode-core E2E) at `npm audit signatures` to complete the trust chain. |

### Items out of scope for security but already raised by the earlier review

The existing `[Critical] Duplicate GitHub Release creation` and `[Important] Tag interpolation mismatch` comments from the functional review both stand and should be fixed before first release, but they are release-flow correctness bugs, not security vulnerabilities. No overlap with the findings above.

### Would-I-block-merge

No. The two medium findings (SHA pinning, `--ignore-scripts`) are best addressed in a follow-up hardening PR once the basic pipeline is proven on the first release. None of the findings expose a concrete exploitable path today — they widen the surface for a future compromise.

**LGTM — ready to merge** from a security perspective, with the medium findings tracked as follow-ups.

dr14-make:
## Review findings addressed — pushed 016c112

Thanks both. Consolidated response:

### Code review findings
| # | Severity | Finding | Status |
|---|---|---|---|
| 1 | critical | softprops-action-gh-release clobbered release-please's curated changelog | ✅ Fixed. Switched to \`append_body: true\` and dropped \`generate_release_notes: true\`. release-please now fully owns the changelog; the workflow only appends the Install + Docker snippets. |
| 2 | important | \`github.ref_name\` (\`v0.2.0\`) vs docker-published tag (\`0.2.0\`) mismatch | ✅ Fixed. Swapped to \`steps.meta.outputs.version\`. |
| 3 | nit | Source maps in tarball | ⏭️ Kept. On a 55.7 kB tarball the ~4× size bloat buys readable prod stacks; worth it. |

### Security review findings
| # | Severity | Finding | Status |
|---|---|---|---|
| 1 | medium | \`@vN\` tags move — risk from compromised action maintainers | ⏭️ Filed as #34 (SHA-pin + Renovate/Dependabot). Real risk but needs a companion dep-update automation; cleanest as its own PR. |
| 2 | medium | \`npm ci\` runs transitive postinstall scripts, can tamper with dist/ pre-publish | ✅ Fixed. Both \`npm ci\` and \`npm publish\` in publish-npm now use \`--ignore-scripts\`. Explicit \`npm run build\` is unaffected. |
| 3 | low | release-please blast-radius | ⏭️ Subsumed by #34 (SHA-pin release-please). |
| 4 | low | First-release bootstrap / partial-release, granular token, 2FA | ⏭️ Ops hygiene (add to release runbook, not a code change). Will capture in the \`## ⚠️ Before the first release\` section of the PR description. |
| 5 | low | Source maps leak relative paths | ⏭️ Confirmed \`../src/…\` relative, no dev-machine fingerprints. Same trade-off as code-review nit #3. |
| 6 | info | Add \`npm audit signatures\` note to README | ⏭️ Will add in the README-polish follow-up that documents the release + install flow. |

### Summary
- 1 critical + 1 important + 1 security-medium fixed in 016c112.
- 1 security-medium deferred with a tracking issue (#34).
- 4 low/informational items either not actionable here or covered by #34.

Ready for round 2.

dr14-make:
## Security re-review of 016c112

Verified against the final state of `.github/workflows/release.yml` and the full PR diff.

### 1. `--ignore-scripts` fix — confirmed correct
Both `npm ci --ignore-scripts` and `npm publish --ignore-scripts --provenance --access public` are in place in the `publish-npm` job. Explicit `npm run build` between them means no functional regression from skipping package scripts. The supply-chain window during which `NODE_AUTH_TOKEN` is present in the environment no longer executes arbitrary transitive postinstall code. Good.

### 2. `NPM_TOKEN` scope — confirmed tight
`NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}` is set under the step-level `env:` of the `Publish to npm` step, not at job or workflow level. `npm ci` and `npm run build` run without the token in env.

### 3. `append_body: true` + `steps.meta.outputs.version` — no new injection surface
`steps.meta.outputs.version` comes from `docker/metadata-action@v6`'s semver normalization. The `on.push.tags: v*.*.*` filter plus `type=semver` pattern mean the value is constrained to a validated semver string — strictly safer than the previous `${{ github.ref_name }}` which took the raw tag. `append_body: true` is pure body-merge behavior in softprops/action-gh-release; no eval surface.

### 4. Issue #34 faithfulness — confirmed
#34 captures every action in the repo, calls out the `release-please-action` blast-radius (`contents: write` + `pull-requests: write` on every push to main), proposes SHA pinning with version comments, and requires Renovate/Dependabot automation so pins don't fossilize. Acceptance criteria match.

### 5. New issues introduced — none blocking
- The new `release-please.yml` uses `@v4` moving tag with write scopes; this is exactly the blast-radius captured in #34, properly deferred.
- `publish-npm` permissions (`contents: read`, `id-token: write`) are correctly minimal for OIDC provenance.
- `publish-docker` and `publish-npm` run in parallel — if docker fails, npm still publishes. Behavioral note, not a security issue.
- No secrets leaked into logs, no `pull_request_target` misuse, no shell interpolation of attacker-controlled values.

---

LGTM — all previous findings resolved or acceptably deferred

dr14-make:
## Re-review of 016c112 — verdict

LGTM — all previous findings resolved.

### Verification details

**1. [Critical] Duplicate Release creation — FIXED**
Verified against upstream `softprops/action-gh-release@v3` source (`src/github.ts:560-567`):
```ts
if (config.input_append_body && workflowBody && existingReleaseBody) {
  body = existingReleaseBody + '\n' + workflowBody;
}
```
- When the Release exists (release-please has already created it): existing body is preserved, workflow body appended. Correct.
- `generate_release_notes: true` was dropped. Correct — otherwise GitHub's auto-generated notes would have overwritten release-please's curated changelog path.

Sequence check: release-please-action uses `repos.createRelease` which creates tag + release body atomically in one API call, then the tag-push event fires `release.yml`. By the time `publish-docker` reaches the softprops step, the Release already exists with its body. No race in practice.

Minor caveat (not blocking): if someone ever pushes a `v*.*.*` tag manually without going through release-please, softprops' `createRelease` path (`src/github.ts:886+`) ignores `append_body` and publishes a release containing *only* Install + Docker snippets. Acceptable edge case given release-please is the only sanctioned path.

**2. [Important] Docker-pull tag format — FIXED**
Verified against `docker/metadata-action` README (semver table, line 511 of README): input tag `v1.2.3` with `pattern={{version}}` yields output `1.2.3` — no `v` prefix. `steps.meta.outputs.version` matches the `{{version}}` pattern which is the first entry in the `tags:` list, so `outputs.version = 0.2.0` for tag `v0.2.0`. This now correctly aligns with the image tag that was actually pushed. Correct.

**3. [Nit] Source maps in tarball — acknowledged, author kept**
Fine. Prod-stack readability trade-off is reasonable.

### New concerns introduced: none blocking

- `publish-docker` and `publish-npm` run in parallel under `needs: ci`. If one succeeds and the other fails you get a partial release (npm publishes are immutable, so re-run requires bumping). Minor operational note, not a bug — leaving at your discretion.
- The leading blank line in the `body:` heredoc produces an extra newline between changelog and Install section when appended. Cosmetic only.

### Supply-chain hardening drive-by (016c112 extras) — good

- `npm ci --ignore-scripts` + `npm publish --ignore-scripts` correctly closes the tamper-dist-before-publish window while `NPM_TOKEN` is in env. Explicit `npm run build` compensates for disabled package scripts. Clean.
- `NPM_TOKEN` correctly scoped to the single publish step, `id-token: write` armed for OIDC provenance. Correct.

Ship it.

PR #42, opened 2026-04-22
fix(broker): scope json() to /_test/deliver so hook bodies aren't eaten

dr14-make:
## Summary

Fix a body-dropping bug in the e2e-mock router (PR #32) surfaced while building the dicode-core#137 testcontainers e2e suite.

## The bug

[buildE2EMockRouter](src/broker/e2e-mock.ts#L71-L88) called \`router.use(json())\` at the top of the router. Because the router is mounted at the app root (required — it has to intercept \`/connect/mock\` before Grant), that \`json()\` middleware ran on every request the router saw, **including unrelated \`/u/:uuid/hooks/*\` webhook forwards**. \`application/json\` bodies were consumed by \`json()\` before the actual forward handler's \`express.raw()\` could read them, so webhook payloads silently arrived empty at connected daemons.

## Prod impact: zero

\`isE2EMockEnabled()\` requires \`DICODE_E2E_MOCK_PROVIDER=1\` **and** fails closed under \`NODE_ENV=production\`. The shipped Dockerfile bakes in \`NODE_ENV=production\`, so a deployed container never mounts the router. Dev environments and the e2e test harness (which explicitly sets \`NODE_ENV=test\`) are the only places the bug can reach.

## Fix

Apply \`json()\` only to the \`/_test/deliver\` route (which actually needs to parse a JSON body). \`/connect/mock\` is a GET so doesn't need it.

## Regression test

\`tests/broker/e2e-mock.test.ts\` gains a \"body-passthrough on unrelated routes\" suite that mounts the e2e-mock router alongside a raw-body POST handler on an unrelated path and asserts an \`application/json\` body flows through intact. Verified the test **fails** without this fix (1 failed / 28 passed after reverting the src change), **passes** with it (132/132 green).

## Test plan

- [x] \`npm run typecheck && npm run lint && npm run format:check && npm run test\` — 132/132 green
- [x] Regression test verified to catch the bug (fails on unpatched src)
- [ ] After merge + release 0.1.2, bump the dicode-core e2e test's \`Dockerfile.relay\` \`FROM\` line to \`0.1.2\` → the suite (issue dicode-core#137) starts passing with realistic \`application/json\` content types

### DIFF
--- a/src/broker/e2e-mock.ts
+++ b/src/broker/e2e-mock.ts
@@ -74,13 +74,17 @@ export function buildE2EMockRouter(
   brokerKey: BrokerSigningKey,
 ): Router {
   const router: Router = makeRouter();
-  router.use(json());
 
   router.get("/connect/mock", (req: Request, res: Response) => {
     handleConnectMock(req, res, sessions);
   });
 
-  router.post("/_test/deliver", (req: Request, res: Response) => {
+  // Scope json() to this specific route. `router.use(json())` would apply
+  // to every request the router sees — and since the router is mounted at
+  // the app root (to run before Grant), it'd consume JSON bodies on
+  // unrelated routes like /u/:uuid/hooks/*, silently dropping webhook
+  // payloads that carry Content-Type: application/json.
+  router.post("/_test/deliver", json(), (req: Request, res: Response) => {
     void handleDeliver(req, res, relay, brokerKey);
   });
 
--- a/tests/broker/e2e-mock.test.ts
+++ b/tests/broker/e2e-mock.test.ts
@@ -485,3 +485,88 @@ describe("mock routes absent when flag unset", () => {
     expect(response.status).toBe(404);
   });
 });
+
+// ---------------------------------------------------------------------------
+// Regression: the e2e-mock router must NOT globally consume JSON bodies
+// ---------------------------------------------------------------------------
+//
+// The router mounts at the app root (so /connect/mock gets intercepted
+// before Grant). Until this guard was added, `router.use(json())` caused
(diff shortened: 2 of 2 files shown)


dr14-make:
## LGTM — ready to merge

**Correctness**
1. `/_test/deliver` still receives parsed JSON — `json()` is passed as per-route middleware before the handler, so `req.body as Partial<DeliverBody>` still works. Confirmed.
2. `/connect/mock` is a GET that only reads `req.query.state` — no body needed. Correct to drop `json()` from it.
3. Verified `e2e-mock.ts` has no other routes; nothing else in the file depends on a parsed body.
4. Test structure is sound: mounts the router at root (mirroring `src/index.ts:119`), pairs it with a downstream raw-body handler (stand-in for `/u/:uuid/hooks/*path` at `src/index.ts:185`), and asserts exact byte length + content round-trip. A regression (returning to `router.use(json())`) would make `seen.length` 0 instead of 17 and fail both assertions.

**Security**
5. No new attack surface. `json()` stays on exactly the one route that needs it; limit/strict defaults unchanged.
6. Prod guard unaffected — `isE2EMockEnabled()` still refuses when `NODE_ENV=production`, so the router never mounts in prod regardless of this change.
7. Test uses ephemeral generated keys and a random listen port; no secrets, no network egress, no fixtures leaked.

Nice catch — the router-wide middleware was a real silent-data-loss bug for JSON webhooks whenever the mock flag is on in dev/test.

*(Note: ran into local git conflict while sanity-checking — unrelated older stash; fix branch content itself is untouched.)*

PR #48, opened 2026-04-23
security(status): use timingSafeEqual for dashboard password compare

### LINKED ISSUE #47: security(status): password compare is timing-unsafe
## Security issue

[\`src/status/auth.ts:28\`](https://github.com/dicode-ayo/dicode-relay/blob/main/src/status/auth.ts#L28) uses short-circuit string comparison:

\`\`\`ts
if (providedPassword !== password) { ... }
\`\`\`

This leaks length and prefix information via timing. Low-bandwidth oracle against the single shared status-dashboard password.

## Fix

Use \`crypto.timingSafeEqual\` on equal-length buffers:

\`\`\`ts
const provided = Buffer.from(providedPassword);
const expected = Buffer.from(password);
if (provided.length !== expected.length || !timingSafeEqual(provided, expected)) { ... }
\`\`\`

Add a test asserting that rejection timing is independent of prefix length (or at minimum a smoke test for the new code path).

## Alpha (dicode-core#82)

Blocker — 30-line fix, no reason to ship without it.

dr14-make:
Closes #47

## Summary
- Replace short-circuit `!==` string compare in `src/status/auth.ts:28` with `crypto.timingSafeEqual` on equal-length buffers.
- Keep an explicit length-mismatch guard before `timingSafeEqual` (which throws on unequal lengths). This leaks only the password length — a fixed configured constant, not secret material.

## Test plan
- [x] `npm run lint` clean
- [x] `npm test tests/status/auth.test.ts` — 6 tests pass, including new cases for wrong-password-same-length and wrong-password-different-length

### DIFF
--- a/src/status/auth.ts
+++ b/src/status/auth.ts
@@ -1,3 +1,5 @@
+import { timingSafeEqual } from "node:crypto";
+
 import type { RequestHandler } from "express";
 
 export function statusAuth(password: string | undefined): RequestHandler {
@@ -25,7 +27,9 @@ export function statusAuth(password: string | undefined): RequestHandler {
     const colonIndex = decoded.indexOf(":");
     const providedPassword = colonIndex === -1 ? decoded : decoded.substring(colonIndex + 1);
 
-    if (providedPassword !== password) {
+    const provided = Buffer.from(providedPassword, "utf8");
+    const expected = Buffer.from(password, "utf8");
+    if (provided.length !== expected.length || !timingSafeEqual(provided, expected)) {
       res.status(401).json({ error: "invalid credentials" });
       return;
     }
--- a/tests/status/auth.test.ts
+++ b/tests/status/auth.test.ts
@@ -68,4 +68,22 @@ describe("statusAuth middleware", () => {
     const body = (await res.json()) as { ok: boolean };
     expect(body.ok).toBe(true);
   });
+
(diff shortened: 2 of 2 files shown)


dr14-make (commented):
LGTM (posting as comment — GitHub blocks self-approve). Targeted, minimal fix.

**Correctness**
- `timingSafeEqual` imported from `node:crypto`, applied to equal-length `Buffer`s encoded consistently as UTF-8 on both sides. `password` is guaranteed defined at the comparison site by the earlier `password === undefined` guard.
- Length-mismatch short-circuit is required (Node's `timingSafeEqual` throws on unequal lengths) and the new test `rejects wrong password of different length without throwing` locks that invariant in.
- Basic-auth header parsing untouched; `decoded.indexOf(":")` fallback preserves pre-existing behavior.

**Security**
- Constant-time byte compare applied correctly. Length-mismatch path does leak password byte length via timing, but that is a fixed deploy-time constant (not per-request-variable secret material), and the PR description calls this out explicitly. Acceptable tradeoff for a single shared dashboard password.
- UTF-8 encoding applied symmetrically — no mixed-encoding pitfalls.
- No regression in header parsing, no new input sinks.

**Tests**
- Equal-length-wrong-password and different-length-wrong-password cases both exercised. Good coverage of the new branch.

Nothing blocking. Nice tight change.

PR #49, opened 2026-04-23
security(broker): allowlist OAuth response fields before encryption

### LINKED ISSUE #44: security(broker): allowlist OAuth response fields before ECIES — don't ship full `req.query`
## Security issue

[\`src/broker/router.ts:207-210\`](https://github.com/dicode-ayo/dicode-relay/blob/main/src/broker/router.ts#L207-L210) builds the token payload from **the full \`req.query\` object**:

\`\`\`ts
const rawTokenData = req.query as Record<string, unknown>;
const tokensToDeliver = Object.fromEntries(
  Object.entries(rawTokenData).filter(([k]) => k !== "state"),
);
\`\`\`

Any attacker-controllable callback-URL parameter (anything Grant leaves through, plus arbitrary \`?x=y\` junk) gets JSON-encoded, ECIES-encrypted, and delivered to the daemon as part of the token record. A malicious upstream or an open redirect exploited upstream could smuggle extra attacker-chosen fields into the daemon's stored secrets.

## Fix

Allowlist the known OAuth response fields and drop everything else before encrypting:

\`\`\`ts
const ALLOWED = ["access_token", "refresh_token", "token_type", "expires_in", "scope", "id_token", "raw"];
const tokensToDeliver = Object.fromEntries(
  Object.entries(rawTokenData).filter(([k]) => ALLOWED.includes(k)),
);
\`\`\`

Add a unit test that a \`?injected=evil\` param in the callback does not reach the encrypted payload.

## Alpha (dicode-core#82)

Blocker — security-sensitive path with a straightforward fix.

dr14-make:
Closes #44

## Summary
Before this change, `handleCallback` in `src/broker/router.ts` forwarded the full `req.query` object (minus `state`) into the ECIES-encrypted token payload delivered to the daemon. Any callback-URL parameter an upstream provider leaves through — or that a malicious upstream / open-redirect injects — would be encrypted alongside the real token and stored in the daemon's secrets.

Now the callback handler explicitly allowlists the OAuth 2.0 RFC 6749 §5.1 response fields plus `id_token` (OIDC) and `raw` (Grant's convenience field with the unmodified provider response body). Everything else is dropped before encryption.

## Threat model
- Attacker-controllable upstream → can append arbitrary query params to the callback URL.
- Open redirect exploited at the provider → same effect.
- Without the allowlist, `?injected=evil` reaches the daemon's stored secrets.
- With the allowlist, only standard token fields survive.

## Implementation
- New exported `ALLOWED_TOKEN_FIELDS` constant + `filterCallbackTokenFields` helper in `src/broker/router.ts`.
- `handleCallback` now calls the helper; the old `filter(([k]) => k !== "state")` is gone.

## Test plan
- [x] New unit test `tests/broker/callback-allowlist.test.ts` (5 cases): standard fields pass through, injected fields dropped, state dropped, empty input, ALLOWED_TOKEN_FIELDS matches the spec.
- [x] `npm run lint` clean

### DIFF
--- a/src/broker/router.ts
+++ b/src/broker/router.ts
@@ -53,9 +53,7 @@ const ALLOWED_TOKEN_FIELDS_SET = new Set<string>(ALLOWED_TOKEN_FIELDS);
 export function filterCallbackTokenFields(
   raw: Record<string, unknown>,
 ): Partial<Record<AllowedTokenField, unknown>> {
-  return Object.fromEntries(
-    Object.entries(raw).filter(([k]) => ALLOWED_TOKEN_FIELDS_SET.has(k)),
-  );
+  return Object.fromEntries(Object.entries(raw).filter(([k]) => ALLOWED_TOKEN_FIELDS_SET.has(k)));
 }
 
 /**
--- a/src/broker/router.ts
+++ b/src/broker/router.ts
@@ -20,6 +20,43 @@ import type { OAuthTokenDeliveryPayload } from "../relay/protocol.js";
 import { buildDeliverySignaturePayload, type BrokerSigningKey } from "./signing.js";
 
 /**
+ * Fields that may be forwarded from an OAuth callback into the encrypted
+ * token delivery payload sent to the daemon.
+ *
+ * Rationale: the callback URL is attacker-reachable — any query param an
+ * upstream provider leaves through (or an open redirect/malicious upstream
+ * injects) would otherwise be JSON-encoded, encrypted to the daemon, and
+ * stored alongside the real token. We allowlist the OAuth 2.0 response
+ * parameters from RFC 6749 §5.1 plus the standard OIDC `id_token` and
+ * Grant's `raw` convenience field (the unmodified provider response body).
+ * Everything else — including `state`, provider error fields, and any
+ * unknown keys — is dropped before encryption.
+ */
+export const ALLOWED_TOKEN_FIELDS = [
+  "access_token",
+  "refresh_token",
+  "token_type",
+  "expires_in",
+  "scope",
+  "id_token",
+  "raw",
+] as const;
+
+type AllowedTokenField = (typeof ALLOWED_TOKEN_FIELDS)[number];
+
+/**
+ * Filter an OAuth callback's query parameters down to allowlisted token fields.
+ * Exported so it can be unit-tested without spinning up an Express server.
+ */
+export function filterCallbackTokenFields(
+  raw: Record<string, unknown>,
+): Partial<Record<AllowedTokenField, unknown>> {
+  return Object.fromEntries(
+    Object.entries(raw).filter(([k]) => (ALLOWED_TOKEN_FIELDS as readonly string[]).includes(k)),
+  ) as Partial<Record<AllowedTokenField, unknown>>;
+}
+
+/**
  * Build the Express router for the OAuth broker.
  *
  * @param relay               RelayServer instance
@@ -202,12 +239,11 @@ async function handleCallback(
     return;
   }
 
-  // Build the token payload (include all grant-returned fields)
-  // Remove state from token data to avoid leaking session metadata
-  const rawTokenData = req.query as Record<string, unknown>;
-  const tokensToDeliver = Object.fromEntries(
-    Object.entries(rawTokenData).filter(([k]) => k !== "state"),
-  );
+  // Build the token payload by allowlisting known OAuth response fields.
+  // The callback URL is attacker-reachable, so we must never ship the full
+  // `req.query` — unknown params (state, session metadata, injected junk)
+  // are dropped before encryption. See ALLOWED_TOKEN_FIELDS above.
+  const tokensToDeliver = filterCallbackTokenFields(req.query as Record<string, unknown>);
 
   // Encrypt the token payload with ECIES for the daemon. The message type
   // is bound into GCM's authenticated data so the daemon can never decrypt
--- a/src/broker/router.ts
+++ b/src/broker/router.ts
@@ -44,6 +44,8 @@ export const ALLOWED_TOKEN_FIELDS = [
 
 type AllowedTokenField = (typeof ALLOWED_TOKEN_FIELDS)[number];
 
+const ALLOWED_TOKEN_FIELDS_SET = new Set<string>(ALLOWED_TOKEN_FIELDS);
+
 /**
  * Filter an OAuth callback's query parameters down to allowlisted token fields.
  * Exported so it can be unit-tested without spinning up an Express server.
@@ -52,8 +54,8 @@ export function filterCallbackTokenFields(
   raw: Record<string, unknown>,
 ): Partial<Record<AllowedTokenField, unknown>> {
   return Object.fromEntries(
-    Object.entries(raw).filter(([k]) => (ALLOWED_TOKEN_FIELDS as readonly string[]).includes(k)),
-  ) as Partial<Record<AllowedTokenField, unknown>>;
+    Object.entries(raw).filter(([k]) => ALLOWED_TOKEN_FIELDS_SET.has(k)),
+  );
 }
 
 /**
@@ -243,7 +245,7 @@ async function handleCallback(
   // The callback URL is attacker-reachable, so we must never ship the full
   // `req.query` — unknown params (state, session metadata, injected junk)
   // are dropped before encryption. See ALLOWED_TOKEN_FIELDS above.
-  const tokensToDeliver = filterCallbackTokenFields(req.query as Record<string, unknown>);
+  const tokensToDeliver = filterCallbackTokenFields(req.query);
 
   // Encrypt the token payload with ECIES for the daemon. The message type
   // is bound into GCM's authenticated data so the daemon can never decrypt
--- a/tests/broker/callback-allowlist.test.ts
+++ b/tests/broker/callback-allowlist.test.ts
@@ -1,0 +1,74 @@
+/**
+ * OAuth callback allowlist — dicode-relay#44.
+ *
+ * The callback URL is attacker-reachable: a malicious upstream OAuth
+ * provider or an open-redirect exploited upstream could append extra
+ * query parameters. Before #44 those extras were JSON-encoded, ECIES-
+ * encrypted, and delivered to the daemon alongside the real token.
+ *
+ * The fix: allowlist the OAuth 2.0 RFC 6749 §5.1 response fields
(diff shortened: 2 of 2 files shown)


dr14-make (commented):
## Verdict: LGTM (comment-only — cannot self-approve)

Clean, minimal fix for #44. The allowlist is correctly placed at the broker→daemon encryption boundary, covers exactly what the Go daemon consumes (`access_token`, `refresh_token`, `scope`, `token_type`, `expires_in` per `dicode-core/pkg/ipc/oauth_store.go:55-116`), and the `id_token` + `raw` additions are well-justified. Good unit-test coverage with explicit negative cases (including a `__proto__` injection attempt).

### Security review

- **Allowlist completeness (vs. daemon consumer):** Verified against `dicode-core/pkg/ipc/oauth_store.go` — the daemon only reads the five RFC 6749 §5.1 fields. Any provider-specific extras (Stripe's `stripe_user_id`, Notion's `workspace_id`, Salesforce's `instance_url`) are not consumed by the broker-delivery path; the `_oauth-app/task.ts` that consumes them is the **direct** non-broker flow (does its own `exchangeCode`). If those are ever wanted in the broker flow later, they're reachable via Grant's `raw` field which is allowlisted. Good.
- **Prototype pollution:** `Object.fromEntries(Object.entries(raw).filter(...))` is safe here — `Object.entries` only enumerates own string keys, and even if `__proto__` appears as an own key from `qs`-parsed queries, the allowlist rejects it. The returned object still has `Object.prototype` as its prototype (not polluted). Confirmed by local repro. The existing test case proves this.
- **Duplicate-key override (`?access_token=real&access_token=evil`):** Not exploitable end-to-end. Express default `qs` parser produces `req.query.access_token = ["real","evil"]`. The allowlist passes arrays through unchanged, so an array lands in the encrypted payload. The Go daemon's `raw[\"access_token\"].(string)` type-assertion at `oauth_store.go:55` fails on non-string, returning \"missing access_token\" — the user sees a failed auth, no secret leak and no override. Note that `handleCallback` already flattens the array to the first element at `router.ts:195-197` for the presence check but doesn't use that flattened value for the payload. Minor nit, not blocking.

### Code review

- `Object.fromEntries(Object.entries(...).filter(...))` is idiomatic and readable. A `for`-loop writing into a fresh `{}` would be marginally faster but not worth it.
- The `as Partial<Record<AllowedTokenField, unknown>>` cast is unavoidable given `Object.fromEntries`'s loose return type; scoped tightly, fine.
- `(ALLOWED_TOKEN_FIELDS as readonly string[]).includes(k)` — the cast is needed because `.includes` on a `readonly` tuple narrows the argument. Clean.

### Nits (non-blocking)
1. Consider tightening the array-duplicate case in a follow-up: `filterCallbackTokenFields` could coerce arrays to `arr[0]` for allowlisted keys, matching the flattening logic already present at `router.ts:195-197`. Current behavior is safe (daemon rejects), but the error surface would be cleaner.
2. The test at `callback-allowlist.test.ts:33` asserts `__proto__` is filtered but doesn't assert the returned object's prototype is still `Object.prototype`. One-line addition would make the pollution guarantee explicit.

Ship it.

dr14-make (commented):
## Verdict

LGTM. The allowlist correctly closes the issue described in #44: `req.query` is no longer piped wholesale into the ECIES-encrypted payload, and only the RFC 6749 §5.1 response fields plus `id_token` (OIDC) and `raw` (Grant) survive. The pure-function split (`filterCallbackTokenFields`) is a clean testability win, and the lint fix on the follow-up commit is appropriately minimal.

## Strengths

- **Scope is tight and defensible.** The 7-field allowlist matches exactly what the daemon-side consumer (`dicode-core` `pkg/ipc/oauth_store.go`) needs; `state`, `error`, `error_description`, and provider-specific junk are all correctly dropped.
- **Threat model is documented in-source.** The JSDoc on `ALLOWED_TOKEN_FIELDS` and the inline comment at the call-site both explain *why* the filtering exists, which matters for a security-sensitive path where future contributors might be tempted to "just add one more field."
- **Test covers the right shapes.** Standard pass-through, attacker injection, `state` leak, empty input, and spec-conformance are all exercised. Exporting the helper rather than reaching through Express is the right call.
- **Set-based lookup** (`ALLOWED_TOKEN_FIELDS_SET`) is O(1) and avoids the `includes()` linear scan from the issue's suggested fix — small win, but free.

## Concerns

None blocking.

One observation, not a blocker: `filterCallbackTokenFields` accepts `Record<string, unknown>` but `req.query` is Express's `ParsedQs`, where values can be `string | string[] | ParsedQs | ParsedQs[]`. A crafted `?raw[evil]=1` would make `raw` a nested object rather than the Grant-provided response body, and that nested object would still pass the allowlist (the filter only gates keys, not value shapes). The daemon then stores whatever JSON lands there. This isn't a regression — the pre-fix code had the same property — and the daemon has its own parsing, so I'd leave it as-is for this PR. Worth a follow-up issue if `raw` is ever consumed structurally downstream.

## Nits

- The `__proto__: "pollution"` entry in the "drops injected attacker-controlled fields" test is a little misleading: JS object-literal semantics mean `__proto__` in a literal sets the prototype rather than an own property, so `Object.entries` never even sees it — the filter isn't actually being exercised on that key. The test still passes (and the outcome is correct: `__proto__` would be dropped), but if you want a true prototype-pollution case, construct the input with `Object.defineProperty(obj, "__proto__", { value: "pollution", enumerable: true })` or use a `null`-prototype object. Low priority.
- `ALLOWED_TOKEN_FIELDS matches RFC 6749 §5.1 + id_token + raw` is effectively a change-detector test — it duplicates the constant. Fine as a tripwire against accidental edits, just noting it's not catching behavioral bugs.

dr14-make (commented):
## Security review — PR #49 (parallel to code-review)

**Verdict:** No P0/P1 security blockers. Safe to merge from a security standpoint. One P2 worth noting about `raw` passthrough semantics, and one P2 defence-in-depth hardening suggestion. The prior self-review already covered prototype pollution / duplicate-key / daemon-consumer completeness — this review focuses on residual attack surface and things that weren't explicitly asserted.

---

### Threat model revisited

The change is placed exactly where it needs to be: the single broker→daemon encryption boundary at `src/broker/router.ts:248`. I grepped the repo for every `eciesEncrypt` call site:

- `router.ts:257` — the path this PR fixes. ✅ allowlisted.
- `e2e-mock.ts:160` — gated by `isE2EMockEnabled()` (fails closed on `NODE_ENV=production` AND requires `DICODE_E2E_MOCK_PROVIDER=1`), and its input is a JSON body from `/_test/deliver`, not `req.query`. Pre-existing dev-only escape hatch, out of scope for #44.
- `tests/broker/crypto.test.ts` — test-only.

No other path builds an encrypted payload from HTTP input. The allowlist is complete for the attacker-reachable surface.

---

### Findings

**P2 — `raw` passthrough is a transitive-trust channel (documented, not a regression)**
`raw` is allowlisted so Grant can ship the unmodified upstream response to the daemon for provider-specific fields. Under Grant's default `transport: 'querystring'`, the entire upstream response body is serialized into `req.query`, and `raw` is itself a JSON string of that body. Consequence: if the daemon ever JSON-decodes `raw` and trusts nested keys (e.g. `raw.admin = true`), the allowlist does not protect those nested fields — they are by construction attacker-influenceable at an OAuth-provider-compromise level.

Per the prior self-review, the Go daemon currently only reads the five top-level RFC 6749 §5.1 fields and does not descend into `raw`, so this is not a live vulnerability. Worth a one-line comment on `ALLOWED_TOKEN_FIELDS` noting "`raw` contents are NOT sanitized — downstream consumers must not trust nested fields from `raw` without their own validation." The rationale block mentions `raw` briefly; making the passthrough caveat explicit would prevent future-self footguns if someone adds a `raw.email_verified` shortcut in the daemon.

**P2 — duplicate-key hardening (defence in depth)**
`?access_token=real&access_token=evil` produces an array that slides through the filter unchanged, and the daemon's `raw["access_token"].(string)` type assertion fails → user sees auth failure. Safe today because the daemon refuses non-strings. Defence-in-depth would be to flatten to the first element for allowlisted keys inside `filterCallbackTokenFields`, matching the flattening already done at `router.ts:222-226` for the presence check. This converts "daemon happens to reject" into "broker normalises" — narrower failure surface and symmetric with the presence-check site. Not blocking.

---

### Notes

- `Object.fromEntries(Object.entries(x).filter(...))` — prototype chain of the returned object is `Object.prototype` with no polluted slots (own `__proto__` key from `qs` is rejected by the allowlist; inherited slots are untouched because `Object.entries` only yields own enumerable string keys). The existing test at `callback-allowlist.test.ts:33` proves the key is filtered; the prior self-review's nit about also asserting the prototype is still `Object.prototype` remains a nice-to-have but is not load-bearing.
- `new Set<string>(ALLOWED_TOKEN_FIELDS)` at module scope: safe. Evaluated once on import, closed-over by the filter function. No mutation path, no ordering concerns, not test-mockable in a way that would escalate.
- Crypto primitives are untouched by this diff — `eciesEncrypt(session.pubkey, session.sessionId, deliveryType, plaintext)` call shape and AAD binding of `deliveryType` are preserved at `router.ts:257`. The `broker_sig` assembly at `router.ts:273-282` is also unchanged. No reordering of session-delete vs. forward that would regress single-use semantics (`sessions.delete` at line 285, before `relay.forward` — unchanged).
- No new logging of sensitive data. `handleCallback` never logs `req.query`, the allowlisted object, or the plaintext. Tests use obvious string literals (`"abc"`, `"def"`, `"real"`), no leaked real credentials.
- `error`, `error_description`, `error_uri` (RFC 6749 §5.2) are dropped by the allowlist, which is correct — the early-return at line 228-230 handles the error path before reaching encryption.

---

### Non-issues (explicitly ruled out)

- **Prototype pollution via `__proto__`**: the allowlist rejects it before `Object.fromEntries` materialises it. Even if it were accepted, `Object.fromEntries` uses `DefineOwnProperty` semantics, so assignment would create an own key on the result rather than traversing the prototype chain. Double-safe.
- **`constructor` / `prototype` keys**: not in the allowlist. Dropped.
- **Confused-deputy via array override**: daemon refuses non-string, user sees failed auth. Not a silent compromise. (See P2 above for a proposed tightening.)
- **Open-redirect via `state`**: `state` is no longer in the encrypted payload, and is still validated against the session store at line 238. No change to that surface.
- **Referrer leak**: `Referrer-Policy: no-referrer` at line 192 is unchanged.
- **Session replay**: `sessions.delete(session.sessionId)` at line 285 unchanged; single-use.

Ship it.

PR #50, opened 2026-04-23
fix(config): honor broker.signing_key_file YAML field

### LINKED ISSUE #46: fix(config): `broker.signing_key_file` YAML field is dead — only env var is read
## Bug

[\`src/config.ts:84\`](https://github.com/dicode-ayo/dicode-relay/blob/main/src/config.ts#L84) declares \`signing_key_file: z.string().default("")\`. [\`relay.yaml:28\`](https://github.com/dicode-ayo/dicode-relay/blob/main/relay.yaml#L28) documents it as a config knob. But a repo-wide grep for \`signing_key_file\` in \`src/\` shows it is **never read**.

[\`src/broker/signing.ts\`](https://github.com/dicode-ayo/dicode-relay/blob/main/src/broker/signing.ts) only branches on env vars \`BROKER_SIGNING_KEY_FILE\` / \`BROKER_SIGNING_KEY\`. An operator who configures via YAML silently gets the auto-generated default.

## Fix

One of:

1. **Wire up the YAML field:** have \`signing.ts\` accept the config value and fall back to env (env takes precedence for ops flexibility).
2. **Remove the dead field:** drop from schema + YAML example + document env-only approach.

Option 1 is preferred — closes the operational footgun without breaking existing env-based deployments.

## Alpha (dicode-core#82)

Blocker — a documented config field that silently does nothing is a trap.

dr14-make:
Closes #46

## Summary
Before this change, `broker.signing_key_file` in `relay.yaml` and `src/config.ts:84` was declared but never read anywhere in `src/`. Operators configuring via YAML silently got the auto-generated `broker-signing-key.pem` in cwd, regardless of what they wrote in the file.

Now `loadBrokerSigningKey(env, cwd, signingKeyFile)` accepts the YAML value as a third parameter. `src/index.ts` passes `brokerCfg.signing_key_file` through.

## Precedence (first hit wins)
1. `BROKER_SIGNING_KEY_FILE` env → PEM from file
2. `BROKER_SIGNING_KEY` env → inline PEM
3. `broker.signing_key_file` YAML → PEM from file (**new**)
4. Auto-generate to `<cwd>/broker-signing-key.pem`

Env still takes precedence so operators can override a baked-in config without re-rendering the file (k8s secret, docker run -e, etc.).

## Test plan
- [x] `npm run lint` clean
- [x] `npm test tests/broker/signing.test.ts` — 9 tests pass, including two new ones: YAML-only path loads the file; env takes precedence over YAML.

### DIFF
--- a/relay.yaml.example
+++ b/relay.yaml.example
@@ -25,7 +25,15 @@ relay:
 
 broker:
   session_ttl_ms: 300000
-  signing_key_file: ""     # auto-generated if empty
+
+  # Broker signing-key file. Precedence: env BROKER_SIGNING_KEY_FILE >
+  # env BROKER_SIGNING_KEY (inline PEM) > this field > auto-generate.
+  # Use a literal path — do NOT use `${ENV_VAR}` interpolation here. If
+  # the env var is unset, interpolation silently collapses to "", skips
+  # this field, and auto-generates a fresh key on every start, rotating
+  # the broker's public key and breaking every daemon that has TOFU-pinned
+  # it. Either write the absolute path literally, or set the env var.
+  signing_key_file: ""     # empty → auto-generate at cwd/broker-signing-key.pem
 
   # OAuth providers — add your own or override defaults.
   # Each key is the Grant provider key and the dicode provider name.
--- a/src/broker/signing.ts
+++ b/src/broker/signing.ts
@@ -2,10 +2,14 @@
  * Broker signing key — proves delivery envelopes were assembled by this
  * relay instance, not by a forger who knows the daemon's public key.
  *
- * Key resolution order:
+ * Key resolution order (first hit wins):
  *   1. BROKER_SIGNING_KEY_FILE env → read PEM from file
  *   2. BROKER_SIGNING_KEY env → inline PEM string
- *   3. Auto-generate to <cwd>/broker-signing-key.pem on first start
+ *   3. broker.signing_key_file from relay.yaml → read PEM from file
+ *   4. Auto-generate to <cwd>/broker-signing-key.pem on first start
+ *
+ * Env takes precedence over YAML so operators can override a baked-in
+ * config without re-rendering the file (e.g. in containers / k8s secrets).
  *
  * The key is ECDSA P-256 (same curve family as the daemon identity, but a
  * completely separate keypair). Only the relay holds the private half; the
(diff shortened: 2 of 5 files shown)


dr14-make (commented):
## Review — fix(config): honor broker.signing_key_file

**Verdict:** approve with minor nits. Precedence chain matches the docstring (env file > env inline > YAML > auto-generate), default `signingKeyFile = ""` is safe, and the sentinel check (`signingKeyFile !== ""`) correctly distinguishes "unset" from "set to a path". Wiring in `src/index.ts` passes `brokerCfg.signing_key_file` through cleanly.

### Correctness
- Precedence is right. Env-overrides-YAML matches the k8s-secrets pattern called out in the new comment block. Good call.
- `readFileSync` on the YAML path will throw synchronously on a bad path — correct fail-fast behavior. Silently falling through to auto-generation would be worse (operator misconfig = a fresh keypair that breaks every daemon's verification).
- No new TOCTOU; the YAML branch does an unconditional `readFileSync`. The existing `existsSync`→`readFileSync` pair in the auto-generate branch is pre-existing and benign (auto-path is inside cwd).
- Auto-gen still writes with `mode: 0o600`.

### Test coverage
Four paths covered (env file / env inline / YAML / auto-gen) and a precedence test — good.

One nit on the precedence test: it asserts that `key.sign`/`verifyDeliverySignature` round-trips, but that would be true for **either** the env pair or the YAML pair (whichever got loaded). It never compares `key.publicKeyBase64` against the env pair's SPKI DER vs the YAML pair's SPKI DER, so it can't actually distinguish "env won" from "YAML won". Suggest:

```ts
const envPubDer = createPublicKey(envPair.privateKey).export({ type: "spki", format: "der" }).toString("base64");
const yamlPubDer = createPublicKey(yamlPair.privateKey).export({ type: "spki", format: "der" }).toString("base64");
expect(key.publicKeyBase64).toBe(envPubDer);
expect(key.publicKeyBase64).not.toBe(yamlPubDer);
```

The `envPubPem.length > 0` / `yamlPubPem.length > 0` asserts at the end are dead weight (and the "neutralize lints" comment is a tell) — they can be dropped once the real comparison above is in.

### Security / operational footgun
`src/config.ts` resolves `${ENV_VAR}` interpolation recursively over **all** string fields via `resolveDeep`, which includes `broker.signing_key_file`. This does not cause a correctness bug — if an operator writes:

```yaml
broker:
  signing_key_file: "${BROKER_SIGNING_KEY_FILE}"
```

...and the env var IS set, the env-file branch wins in `loadBrokerSigningKey` before the YAML path is even consulted, so there's no shadowing. If the env var is UNSET, the interpolation collapses to `""`, the YAML branch is skipped (due to `signingKeyFile !== ""`), and we fall through to **auto-generation of a brand-new key**. On a restart this silently rotates the broker pubkey — every connected daemon would reject signatures until they re-pull the welcome message. That's a real foot-gun if someone templates the YAML expecting "unset env → use the YAML literal."

Not a blocker for this PR (the behavior predates it), but worth either:
1. Logging a warning when `signing_key_file` resolves to `""` after interpolation (distinguishable from "field omitted entirely" by checking the raw YAML before `resolveDeep`), or
2. A docs note on `relay.yaml.example` that `signing_key_file` should be a literal path, not a `${...}` reference (use the env var directly instead).

Prefer option 2 — cheapest and matches the "env takes precedence" model already documented.

### Nits
- Docstring on line 46 says `signing_key_file` from relay.yaml "(or \"\" if unset)" — accurate, but worth noting that the Zod default is what guarantees the empty-string sentinel at the call site. A reader tracing the invariant will need to hop to `BrokerSchema` in `src/config.ts`. Consider a one-line pointer.

PR #51, opened 2026-04-23
docs(readme): update handshake examples to post-#104 split-key schema

### LINKED ISSUE #45: docs(readme): handshake example is pre-#104 (missing `decrypt_pubkey`)
## Docs drift

[\`README.md:130-137\`](https://github.com/dicode-ayo/dicode-relay/blob/main/README.md#L130-L137) shows a \`hello\` example with only \`type\`, \`uuid\`, \`pubkey\`, \`sig\`, \`timestamp\` — no \`decrypt_pubkey\`.

The implementation at [\`src/relay/protocol.ts:72\`](https://github.com/dicode-ayo/dicode-relay/blob/main/src/relay/protocol.ts#L72) now requires \`decrypt_pubkey: z.string()\` (not optional), and [\`src/relay/server.ts:392-406\`](https://github.com/dicode-ayo/dicode-relay/blob/main/src/relay/server.ts#L392-L406) hard-rejects a missing/malformed one.

Any integrator mirroring the README verbatim produces an invalid handshake.

Validator also flagged that the \`welcome\` example at README:139-144 is missing \`protocol\` and \`broker_pubkey\` — even more stale.

## Fix

- Update the \`hello\` example to include \`decrypt_pubkey\` (post-#104 split identity).
- Update the \`welcome\` example to include \`protocol: 2\` and \`broker_pubkey\`.

## Alpha (dicode-core#82)

Blocker — integrators following the README cannot connect.

dr14-make:
Closes #45

## Summary
- The `hello` example in `README.md:130-137` was missing `decrypt_pubkey`, which became required at `src/relay/protocol.ts:72` in dicode-core#104. Any integrator mirroring the README verbatim had their handshake hard-rejected at `src/relay/server.ts:392-406`.
- The `welcome` example at `README.md:139-144` was also stale — missing the `protocol: 2` gate and `broker_pubkey` TOFU pin.
- Also updated the security-model paragraph that claimed tokens are encrypted with "the daemon's public key" — now correctly describes the split sign/decrypt identity.

## Test plan
- [x] Manually cross-checked examples against `src/relay/protocol.ts` Zod schemas and `src/relay/server.ts` welcome emission.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,15 +129,21 @@ Server → Client:
 
 Client → Server:
   {
-    "type":      "hello",
-    "uuid":      "<64 lowercase hex>",   // hex(sha256(uncompressed_pubkey))
-    "pubkey":    "<base64 std>",         // 65 bytes: 0x04 || X || Y
-    "sig":       "<base64 std>",         // ECDSA P-256 ASN.1 DER over sha256(nonce_bytes || timestamp_be_uint64)
-    "timestamp": <unix seconds integer>
+    "type":           "hello",
+    "uuid":           "<64 lowercase hex>",   // hex(sha256(uncompressed_pubkey))
+    "pubkey":         "<base64 std>",         // 65 bytes: 0x04 || X || Y — ECDSA signing key
+    "decrypt_pubkey": "<base64 std>",         // 65 bytes: 0x04 || X || Y — ECIES recipient (OAuth token delivery)
+    "sig":            "<base64 std>",         // ECDSA P-256 ASN.1 DER over sha256(nonce_bytes || timestamp_be_uint64)
+    "timestamp":      <unix seconds integer>
   }
 
 Server → Client (success):
-  { "type": "welcome", "url": "wss://relay.dicode.app/u/<uuid>/hooks/" }
+  {
+    "type":          "welcome",
+    "url":           "wss://relay.dicode.app/u/<uuid>/hooks/",
+    "protocol":      2,                        // broker advertises split sign/decrypt key support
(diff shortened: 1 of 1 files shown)


dr14-make (commented):
LGTM — would approve, but GitHub blocks self-approval on own PR. Posting as comment review instead.

Cross-checked the updated examples against the source of truth:
- `hello` example matches `HelloMessageSchema` at `src/relay/protocol.ts:59-77` (all six fields present; `decrypt_pubkey` is required, not optional — confirming the bug this PR fixes).
- `welcome` example matches `WelcomeMessageSchema` at `src/relay/protocol.ts:21-33` and the emission path at `src/relay/server.ts:311-316` (`protocol: PROTOCOL_VERSION` = 2 is always sent; `broker_pubkey` is present whenever configured).
- "base64 SPKI DER" comment on `broker_pubkey` is correct — `src/broker/signing.ts:73-74` does `publicKey.export({ type: "spki", format: "der" }).toString("base64")`.
- Security-model rewording accurately matches `src/broker/router.ts:130-136` (session stores `client.decryptPubkey` as the ECIES recipient) and `src/broker/router.ts:219` (`eciesEncrypt(session.pubkey, …)`).

Non-blocking nit: the `decrypt_pubkey` inline comment describes only the wire encoding (`0x04 || X || Y`); `verifyHello` at `src/relay/server.ts:389-405` additionally enforces on-curve P-256 validation. Fine to leave — the `pubkey` comment right above has the same shape, so symmetry wins over completeness here.

No remaining stale pre-#104 handshake mentions in the README.

PR #56, opened 2026-04-24
refactor(status,broker): adopt express-basic-auth and lru-cache

dr14-make:
## Summary
- Replace hand-rolled basic-auth in `src/status/auth.ts` with `express-basic-auth` — adds timing-safe compare (small security win; current code uses `!==`). Closes item (9) in dicode-ayo/dicode-core#195.
- Replace custom Map+setTimeout TTL in `src/broker/sessions.ts` with `lru-cache`. Adds a hard 10_000-entry ceiling (LRU eviction) to bound memory against abusive traffic. Closes item (10) in dicode-ayo/dicode-core#195.

~83 LOC removed. All existing tests pass unchanged.

### Notes
- `src/status/auth.ts`: 404-when-unset is handled as a thin wrapper before the middleware is installed, since `express-basic-auth` otherwise returns 401 unconditionally.
- `src/broker/sessions.ts`: `LRUCache` is configured with `ttlAutopurge: true` plus a `Date.now()`-backed `perf` clock so fake-timer tests stay deterministic (LRUCache's module-level `performance` reference misses the object Vitest swaps in via `vi.useFakeTimers()`).

## Test plan
- [ ] `tests/status/auth.test.ts` all 4 tests pass
- [ ] `tests/broker/sessions.test.ts` all 8 tests pass
- [ ] `npm run lint` clean
- [ ] `npm run typecheck` clean

### DIFF
--- a/src/broker/sessions.ts
+++ b/src/broker/sessions.ts
@@ -3,8 +3,12 @@
  * Each session is created when the daemon initiates an OAuth flow and is
  * deleted immediately after the token is delivered.
  *
- * TTL: 5 minutes. Sessions that have not been completed by then are purged.
+ * TTL: configurable per instance (5 minutes in production). Sessions that
+ * have not been completed by then are purged. Storage is an `lru-cache` with
+ * a hard max of 10_000 entries to bound memory against abusive traffic.
  */
+
+import { LRUCache } from "lru-cache";
 
 export interface Session {
   /** UUID v4 of the broker session */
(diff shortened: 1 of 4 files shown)


src/broker/sessions.ts
```diff
@@ -23,69 +27,76 @@ export interface Session {
...
   constructor(ttlMs: number) {
-    this.ttlMs = ttlMs;
+    const opts: LruOpts = {
+      max: MAX_SESSIONS,
+      ttl: ttlMs,
+      // Schedule a setTimeout per entry so expired sessions are dropped
+      // eagerly (preserves prior Map+setTimeout behavior: `size` drops to 0
+      // at TTL without needing a subsequent `get`).
+      ttlAutopurge: true,
+      perf: dateClock,
```
dr14-make, line 64: Nit: undocumented API surface. The perf option is accepted by the LRUCache constructor at runtime but isn't part of the published LRUCache.Options type (hence the LruOpts intersection just below). If lru-cache renames/removes this internal in a future minor, the code silently falls back to captured performance, which breaks fake-timer tests (exactly the problem this is meant to solve). A defensive test already exists in tests/broker/sessions.test.ts and will catch regressions, so this is acceptable as-is. Pinning lru-cache to ^11 is sufficient.

src/broker/sessions.ts
```diff
@@ -23,69 +27,76 @@ export interface Session {
   scope?: string | undefined;
 }
 
+/** Hard ceiling on concurrent in-flight OAuth sessions. */
+const MAX_SESSIONS = 10_000;
```
dr14-make, line 31: Consider routing MAX_SESSIONS through the Zod config schema (src/config.ts) the same way session_ttl_ms is. The rest of this repo keeps tunables out of module scope so operators can override via env / config.yaml without a source edit. Non-blocking; happy to address in a follow-up.

src/status/auth.ts
```diff
@@ -1,39 +1,26 @@
...
-    if (provided.length !== expected.length || !timingSafeEqual(provided, expected)) {
-      res.status(401).json({ error: "invalid credentials" });
-      return;
-    }
+    };
+  }
 
-    next();
-  };
+  return basicAuth({
```
dr14-make, line 21: Observable behavior change: previous implementation returned JSON bodies like { error: "authentication required" } / { error: "invalid credentials" } on 401. express-basic-auth defaults to empty string body. Existing tests only assert status code + WWW-Authenticate header so nothing fails, but any external client that parsed the JSON error will now see an empty body. Low risk (this is admin-only status). If you want to keep the JSON, pass unauthorizedResponse: () => ({ error: 'authentication required' }).

PR #58, opened 2026-04-25
chore: dedupe crypto/HTML helpers and adopt lru-cache for nonces

dr14-make:
## Summary
Continuation of the consolidation pass started in #195. Three small behavior-preserving refactors plus a memory bound on the nonce store.

1. **NonceStore → `lru-cache`**. Replaces hand-rolled `Map` + per-entry `setTimeout` with `LRUCache(max=100_000, ttlAutopurge)`. The original `Map` had no ceiling — a flood of fresh nonces against the handshake endpoint would grow memory unbounded. Same `dateClock` pattern as `SessionStore` (#195) so `vi.useFakeTimers()` tests stay deterministic. Public API (`check` / `size` / `clear`) unchanged.

2. **`uncompressedP256ToSpki` extracted to `src/shared/crypto-utils.ts`**. The fixed 27-byte SPKI prefix for `ecPublicKey + prime256v1` was duplicated verbatim in `src/relay/server.ts` and `src/broker/crypto.ts`. Crypto duplication is a footgun — one shared module, both call sites import from it. Deliberately *not* replaced with an external package: the byte layout is part of the wire contract with the Go daemon.

3. **`escape-html` for HTML escaping**. Two hand-rolled `escape*` functions collapsed into the Express-blessed `escape-html` package. The previous `esc()` in `page.ts` didn't escape single quotes; `escape-html` does — strict superset, no XSS regression. `@types/escape-html` added to devDependencies.

Duration formatting in `metrics.ts`/`page.ts` left as-is — deduping risks output drift, lowest-priority item from the research.

## Why now
Filed from an open-source-replacement research pass (2026-04-25). Companion to dicode-core#203 (cenkalti/backoff) and dicode-core#204 (bcrypt passphrase hashing). The Tink ECIES question is deferred to dicode-core#202.

## Test plan
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run test` (144/144)
- [x] `npm run build`
- [x] `npm run format:check`
- [ ] Manual: confirm a flood of unique nonces is bounded at 100k entries (visible via `NonceStore.size`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
   },
   "dependencies": {
     "@bufbuild/protobuf": "^2.12.0",
+    "escape-html": "^1.0.3",
     "express": "^5.0.1",
     "express-basic-auth": "^1.2.1",
     "grant": "^5.4.22",
@@ -67,6 +68,7 @@
   "devDependencies": {
     "@bufbuild/protoc-gen-es": "^2.12.0",
     "@eslint/js": "^10.0.1",
+    "@types/escape-html": "^1.0.4",
     "@types/express": "^5.0.0",
     "@types/js-yaml": "^4.0.9",
     "@types/node": "^25.6.0",
--- a/src/relay/nonces.ts
+++ b/src/relay/nonces.ts
@@ -1,19 +1,44 @@
 /**
- * NonceStore — prevents replay attacks by tracking seen nonces for 60 seconds.
- * O(1) lookup and insertion. Entries are evicted after their TTL expires.
+ * NonceStore — prevents replay attacks by tracking seen nonces for the
+ * configured TTL window (60 s in production). Backed by `lru-cache` so
+ * that an attacker flooding the handshake endpoint with fresh nonces
+ * cannot grow memory unbounded — entries are LRU-evicted at the ceiling.
  */
 
-interface NonceEntry {
-  expiresAt: number;
-  timer: ReturnType<typeof setTimeout>;
-}
+import { LRUCache } from "lru-cache";
+
+/** Hard ceiling on tracked nonces. ~10x the expected steady-state churn. */
+const MAX_NONCES = 100_000;
+
+/**
+ * `Date.now()`-backed clock used by LRUCache for TTL bookkeeping. Same
+ * rationale as SessionStore: `vi.useFakeTimers()` patches the `Date`
+ * global in place but *replaces* `performance` with a fake object, which
+ * LRUCache's module-level `defaultPerf` reference would miss. Real-world
+ * drift from system clock adjustments is acceptable for a 60 s anti-replay
+ * window.
+ */
+const dateClock = {
+  now: (): number => Date.now(),
+};
+
+type LruOpts = LRUCache.Options<string, true, unknown> & {
+  perf?: { now(): number };
+};
 
 export class NonceStore {
-  private readonly store = new Map<string, NonceEntry>();
-  private readonly ttlMs: number;
+  private readonly cache: LRUCache<string, true>;
 
   constructor(ttlMs: number) {
-    this.ttlMs = ttlMs;
+    const opts: LruOpts = {
+      max: MAX_NONCES,
+      ttl: ttlMs,
+      // Eager eviction so `size` drops as entries expire — matches the
+      // prior Map+setTimeout behavior the tests assert.
+      ttlAutopurge: true,
+      perf: dateClock,
+    };
+    this.cache = new LRUCache<string, true>(opts);
   }
 
   /**
@@ -21,29 +46,20 @@ export class NonceStore {
    * Registers the nonce so future calls return true.
    */
   check(nonce: string): boolean {
-    if (this.store.has(nonce)) {
+    if (this.cache.has(nonce)) {
       return true;
     }
-    const timer = setTimeout(() => {
-      this.store.delete(nonce);
-    }, this.ttlMs);
-    // Allow Node.js to exit even if the timer is still pending
-    timer.unref();
-
-    this.store.set(nonce, { expiresAt: Date.now() + this.ttlMs, timer });
+    this.cache.set(nonce, true);
     return false;
   }
 
   /** Number of nonces currently tracked (for testing / observability). */
   get size(): number {
-    return this.store.size;
+    return this.cache.size;
   }
 
   /** Clear all stored nonces (for testing). */
   clear(): void {
-    for (const entry of this.store.values()) {
-      clearTimeout(entry.timer);
-    }
-    this.store.clear();
+    this.cache.clear();
   }
 }
--- a/src/relay/nonces.ts
+++ b/src/relay/nonces.ts
@@ -3,6 +3,17 @@
  * configured TTL window (60 s in production). Backed by `lru-cache` so
  * that an attacker flooding the handshake endpoint with fresh nonces
  * cannot grow memory unbounded — entries are LRU-evicted at the ceiling.
+ *
+ * Replay-via-eviction: at sustained handshake rates above
+ * MAX_NONCES / ttlSeconds (≈ 1,667 handshakes/sec for 100k/60s) the LRU
+ * starts evicting *non-expired* nonces, which would in principle allow
+ * an attacker to replay an evicted but still-time-valid nonce. This is
+ * not exploitable in practice because the handshake also enforces a
+ * ±30s timestamp window (see relay protocol spec, verified in
+ * src/relay/server.ts) — captured handshakes whose nonces could be
+ * evicted are already too old on the timestamp check. The 60s nonce
+ * TTL is the inner defense; the timestamp window is the binding outer
+ * bound on replay validity.
  */
 
 import { LRUCache } from "lru-cache";
@@ -44,6 +55,13 @@ export class NonceStore {
   /**
    * Returns true if the nonce has been seen within the TTL window.
    * Registers the nonce so future calls return true.
+   *
+   * Atomicity: the has()/set() pair is *not* a CAS but is safe here
+   * because Node.js JavaScript runs on a single thread and this method
+   * contains no `await` — no other handshake can run between the two
+   * operations. If a future contributor adds an `await` mid-method,
+   * concurrent handshakes with the same nonce could both see has()
+   * return false and both proceed, which is a replay-acceptance bug.
    */
   check(nonce: string): boolean {
     if (this.cache.has(nonce)) {
--- a/src/shared/crypto-utils.ts
+++ b/src/shared/crypto-utils.ts
@@ -1,0 +1,26 @@
+/**
+ * Crypto helpers shared between the relay handshake (`src/relay/server.ts`)
+ * and the OAuth broker (`src/broker/crypto.ts`).
+ *
+ * Kept in a shared module rather than a runtime dependency: the DER wrapping
+ * is a fixed 27-byte SPKI prefix specific to ecPublicKey + prime256v1, and
+ * we want exact control over the byte layout the daemon expects on the wire.
+ */
+
+import { Buffer } from "node:buffer";
+
+/**
+ * Fixed SPKI header for ecPublicKey + prime256v1 (P-256). When concatenated
+ * with a 65-byte uncompressed public key (`0x04 || X || Y`) this yields a
+ * DER-encoded SubjectPublicKeyInfo Node's `node:crypto` can import.
+ */
+const P256_SPKI_HEADER = Buffer.from("3059301306072a8648ce3d020106082a8648ce3d030107034200", "hex");
+
+/**
+ * Wraps a raw 65-byte uncompressed P-256 public key into a DER-encoded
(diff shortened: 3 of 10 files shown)


src/relay/nonces.ts
```diff
@@ -1,49 +1,65 @@
...
     }
-    const timer = setTimeout(() => {
-      this.store.delete(nonce);
-    }, this.ttlMs);
-    // Allow Node.js to exit even if the timer is still pending
-    timer.unref();
-
-    this.store.set(nonce, { expiresAt: Date.now() + this.ttlMs, timer });
+    this.cache.set(nonce, true);
     return false;
```
dr14-make, line 71: **TOCTOU between has() and set() — currently safe, but worth a comment.**

The check() body does cache.has(nonce) then cache.set(nonce, true) as two separate operations. In a multi-threaded environment that would be a classic TOCTOU race (two concurrent calls both see has() return false, both set, both return false → replay accepted).

Node is single-threaded and check() contains no await, so the two calls execute atomically within one tick — the race cannot actually happen here. Still, this is load-bearing security code and the atomicity guarantee is implicit. Two suggestions:

1. Add an inline comment like `// Single-tick: no await between has() and set(), so atomic.` so a future contributor doesn't introduce an await mid-method.
2. If async work ever lands in verifyHello before this call (e.g. async crypto verify), keep the nonce check as the *first* gate after the synchronous parsing — moving it after an await would open an actual race against itself if the same nonce is replayed twice in flight.

Non-blocking — flagging because the protocol depends on this being atomic.

src/relay/nonces.ts
```diff
@@ -1,49 +1,65 @@
...
+    this.cache = new LRUCache<string, true>(opts);
   }
 
   /**
    * Returns true if the nonce has been seen within the TTL window.
    * Registers the nonce so future calls return true.
    */
   check(nonce: string): boolean {
-    if (this.store.has(nonce)) {
+    if (this.cache.has(nonce)) {
```
dr14-make, line 67: **Security review — replay via LRU eviction**: Considered. With `max=100_000` and `ttl=60s`, an attacker would need to push >100k unique nonces through `check()` within a single 60s window to evict a still-valid victim nonce — ~1,667 successful handshake initiations/sec sustained. However, even if eviction succeeded, the handshake's `timestamp` window of ±30s (per protocol spec, verified in `server.ts`) is the binding replay constraint, not the 60s nonce TTL: a captured handshake older than ~30s is rejected on timestamp grounds regardless of nonce-store state. The LRU ceiling does not widen the practical replay window. **No replay regression vs. the prior unbounded Map.** (DoS/memory exhaustion is the concern this change actually addresses, and is out of scope for this review.)

**TOCTOU `cache.has()` → `cache.set()`**: Considered. `check()` is fully synchronous with no `await` between the lookup and insertion, and Node.js executes JS on a single thread. Two concurrent handshakes cannot interleave inside this method — each `check()` call runs to completion atomically with respect to other JS. **Not exploitable.**

src/shared/crypto-utils.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+import { Buffer } from "node:buffer";
+
+/**
+ * Fixed SPKI header for ecPublicKey + prime256v1 (P-256). When concatenated
+ * with a 65-byte uncompressed public key (`0x04 || X || Y`) this yields a
+ * DER-encoded SubjectPublicKeyInfo Node's `node:crypto` can import.
+ */
+const P256_SPKI_HEADER = Buffer.from("3059301306072a8648ce3d020106082a8648ce3d030107034200", "hex");
```
dr14-make, line 18: **Security review — SPKI byte equality**: Verified. The hex string `3059301306072a8648ce3d020106082a8648ce3d030107034200` in this shared module is byte-for-byte identical to the two originals deleted from `src/broker/crypto.ts` and `src/relay/server.ts` (compared against `origin/main`: same 52 hex chars / 26 bytes — note the comment says 27 but it's actually 26 bytes; `Buffer.concat([header(26), pubkey(65)])` = 91-byte SPKI, which is the correct DER length for ecPublicKey + prime256v1 with an uncompressed point). **No transcription error introduced by the move.** Minor doc nit (header is 26 bytes, not 27) is non-security.

src/relay/nonces.ts
```diff
@@ -1,49 +1,65 @@
...
  */
 
-interface NonceEntry {
-  expiresAt: number;
-  timer: ReturnType<typeof setTimeout>;
-}
+import { LRUCache } from "lru-cache";
+
+/** Hard ceiling on tracked nonces. ~10x the expected steady-state churn. */
+const MAX_NONCES = 100_000;
```
dr14-make, line 22: **100,000 ceiling — flagging the LRU-eviction replay window.**

The defense changes shape with this PR. Before: every nonce was held for the full 60 s TTL. After: at sustained handshake rates above (MAX_NONCES / ttlSeconds) = 100_000 / 60 ≈ 1,667 handshakes/sec, the LRU starts evicting *non-expired* nonces. An attacker who can replay an evicted nonce within its original 60-s validity window defeats the anti-replay guard.

Practical assessment:
- Steady-state replay protection still holds at any rate well below ~1,600 hps. The PR comment says "~10x the expected steady-state churn" — that's fine if the expected churn is ~150 hps, but there's no source for that number in the PR description.
- An attacker can't easily *cause* eviction of a specific target nonce via flooding because LRU evicts in insertion order, not by content. So flooding mostly protects the attacker's *own* old nonces, not a specific victim's. That softens the threat considerably.
- However, the timestamp ±30 s check is the actual outer bound on replay (line 446 area in server.ts). Nonces expiring at 60 s vs evicting at <60 s under load only matters if an attacker can also forge a fresh-looking timestamp, which they can't (signature covers the timestamp).

Net: probably fine, but please:
1. Document the trade-off explicitly in the file header — "LRU eviction can drop entries before TTL under flood; the timestamp ±30 s window is the real outer replay bound."
2. Consider exposing `size` via a metric so operators can see when steady-state nears the ceiling.
3. Sanity-check 100k vs the actual handshake rate you've measured — if real traffic is 10 hps the ceiling could be 10x lower; if it's 1k hps the ceiling is uncomfortably close.

package.json
```diff
@@ -55,6 +55,7 @@
   },
   "dependencies": {
     "@bufbuild/protobuf": "^2.12.0",
+    "escape-html": "^1.0.3",
```
dr14-make, line 58: **Security review — escape-html@1.0.3 CVE check**: No CVEs or GHSA advisories published against `escape-html` 1.0.3 (latest published, used transitively by Express/serve-static for >a decade). Source inspected locally (`node_modules/escape-html/index.js`): regex `/[\"'&<>]/` and switch-based replacement to `&quot; &amp; &#39; &lt; &gt;` — strict superset of the prior hand-rolled `esc()` in `page.ts` (which omitted single-quote escaping), so the swap is an XSS-tightening change, not a regression. **No issue.**

src/status/page.ts
```diff
@@ -1,3 +1,4 @@
+import escapeHtml from "escape-html";
```
dr14-make, line 1: **escape-html parity check — verified, with one note.**

Confirmed against `node_modules/escape-html/index.js`: the package escapes `& < > " '` (matching against `/["'&<>]/`). That is a strict superset of:
- The old `esc()` in `page.ts` (which escaped `& < > "` only — single quote was missing as the PR description correctly notes).
- The old `escapeHtml()` in `router.ts` (which escaped all five).

So substituting escape-html for both is a parity-preserving (or strictly safer) swap. No XSS regression; the page.ts case is a tiny improvement.

One nit: `escape-html` emits `&#39;` for the single quote, while the old `router.ts` helper emitted `&#x27;`. Both decode identically — semantically equivalent — but if any snapshot/golden tests assert exact HTML output they may need updating. (Skimmed `tests/status/page.test.ts` and `tests/broker/auth.test.ts` and didn't see a strict-equality fixture, so this is a heads-up rather than a known break.)

Also worth confirming explicitly: escape-html does *not* escape characters above U+007E. None of the call sites here emit raw user input into a context where multi-byte UTF-8 could be misinterpreted (they all live inside HTML element bodies/attributes after `</td>`-style markup), so this is fine, but worth keeping in mind if these helpers ever get repurposed for, say, JavaScript-string contexts.

src/shared/crypto-utils.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Wraps a raw 65-byte uncompressed P-256 public key into a DER-encoded
+ * SubjectPublicKeyInfo so it can be passed to Node.js `crypto.createVerify`,
+ * `crypto.createPublicKey`, and friends.
+ */
+export function uncompressedP256ToSpki(pubkey: Buffer): Buffer {
```
dr14-make, line 25: **`src/shared/` placement — fits the codebase, but lock it down before it becomes a junk drawer.**

Checked against the layout in CLAUDE.md: the documented top-level dirs are `src/relay/` and `src/broker/`, with no `src/shared/` or `src/common/`. `src/shared/` is a reasonable choice here — `uncompressedP256ToSpki` truly belongs to neither subsystem and is invoked from both — and CLAUDE.md doesn't explicitly forbid new top-level dirs.

Two suggestions to keep this from sprawling:

1. **Update `CLAUDE.md`'s repo-layout section** so the next agent (or human) knows `src/shared/` is sanctioned and what's allowed there. Without that, a future cleanup pass might "dedupe" things back into `src/relay/` or invent yet another sibling like `src/lib/` or `src/util/`.

2. **Define a scope rule.** Suggested: "`src/shared/` holds protocol-byte-layout utilities used by both relay and broker — anything else stays in its owning subsystem." That keeps `shared/` from becoming a generic util dump (cf. lodash anti-pattern called out in CLAUDE.md). The current 27-byte SPKI prefix is exactly this kind of utility.

Alternative considered: putting `uncompressedP256ToSpki` in `src/broker/crypto.ts` and importing from there. Rejected — would create a relay → broker dep, which inverts the natural layering (relay is the entry point, broker is downstream of it).

`src/shared/` is the right call. Just please document it.

src/relay/nonces.ts
```diff
@@ -1,49 +1,65 @@
...
+    const opts: LruOpts = {
+      max: MAX_NONCES,
+      ttl: ttlMs,
+      // Eager eviction so `size` drops as entries expire — matches the
+      // prior Map+setTimeout behavior the tests assert.
+      ttlAutopurge: true,
+      perf: dateClock,
+    };
+    this.cache = new LRUCache<string, true>(opts);
   }
```
dr14-make, line 53: **Test coverage gap: no test exercises the new memory bound.**

Read `tests/relay/handshake.test.ts` — it covers `check` first/second-call semantics (lines 314-322), independence between nonces, and TTL eviction with fake timers (lines 339-356). That's the existing happy-path; it's preserved correctly.

The new behaviors introduced by *this PR* aren't covered:

1. **MAX_NONCES ceiling.** A test that inserts 100,001 unique nonces and asserts `size === 100_000` would lock the ceiling in. Without it, a future "clean up the magic number" refactor could silently raise/lower 100k and no test would catch the regression.

2. **LRU eviction-before-TTL semantics.** A test that inserts 100,001 nonces and confirms the *first* nonce returns `false` (not `true`) on a re-check would document the replay-via-eviction window I called out separately. It also enshrines that this is *expected* — anyone who later swaps the ceiling for a non-LRU bound would have to update the test deliberately.

3. **`dateClock` plumbing under `vi.useFakeTimers()`.** The existing TTL test (line 339) implicitly verifies this since it relies on `Date.now()`-driven eviction working under fake timers. Good — but worth an inline comment in the test calling out that this is *also* the regression test for the `perf: dateClock` wiring, otherwise someone might "simplify" `dateClock` away thinking it's unused.

The PR description's manual checklist item — "confirm a flood of unique nonces is bounded at 100k entries (visible via NonceStore.size)" — is exactly the test that should be automated. It's a 5-line vitest case.

dr14-make:
Closing without merge — the diff has drifted too far from main to rebase cleanly, but the underlying improvements are still valuable. Refiling as smaller PRs against current main:

**Why close instead of rebase**
- `src/broker/crypto.ts` was deleted (broker restructured post-#56), so the SPKI extraction hunk no longer applies.
- `src/shared/` was introduced organically since this PR was filed and already contains `crypto.ts` / `protocol.ts` / `signing.ts` — the new `crypto-utils.ts` file proposed here would be redundant with the existing `shared/crypto.ts`.
- CLAUDE.md tree changed; `package-lock.json` conflicts on every dependabot bump (#62, #68, #69).
- Net: rebase work ≈ refile work, and the diff is cleaner against current shape.

**Replacement PRs**
1. NonceStore → `lru-cache` with 100k ceiling (highest value — bounds memory; mirrors what `SessionStore` already does).
2. HTML escaping → `escape-html` package, applied to both `broker/router.ts` and `status/page.ts` (also fixes the missing `'` escape).
3. SPKI dedupe — fold `uncompressedP256ToSpki` from `src/relay/server.ts` into existing `src/shared/crypto.ts`.

Will link the new PRs here once filed.

PR #57, opened 2026-04-24
refactor(relay): generate protocol types from protobuf schema

dr14-make:
## Summary

Closes the TypeScript half of item (3) in dicode-ayo/dicode-core#195. Paired with dicode-ayo/dicode-core#199 which does the Go side.

The hand-maintained Zod schema in \`src/relay/protocol.ts\` is replaced by types generated from dicode-core's \`proto/relay.proto\` (vendored here verbatim — must stay in lockstep with dicode-core releases). \`buf generate\` emits \`src/relay/pb/relay_pb.ts\` via \`@bufbuild/protoc-gen-es\`.

## Wire format (protocol version 3 — breaking)

- **headers**: previously \`{\"Content-Type\": [\"application/json\"]}\`, now \`{\"Content-Type\": {\"values\": [\"application/json\"]}}\`. Proto3 maps cannot hold repeated values directly; \`HeaderValues\` wraps the list.
- **timestamp**: \`int32\` on the wire so \`toJson\` emits a JSON number instead of int64's default quoted-string encoding.
- **Envelope**: \`{\"<variant>\": {...}}\` oneof form replaces flat \`{\"type\": \"...\", ...}\`.

\`PROTOCOL_VERSION\` is now \`3\`; Go daemons at dicode-core#199 refuse brokers advertising \`< 3\`.

## Scope note

\`OAuthTokenDeliveryPayload\` (the body of the \`/hooks/oauth-complete\` webhook) stays hand-typed on both sides. It's not part of the tunnel protocol — it's transported inside \`Request.body\` as base64. Keeping it out of the proto matches the Go daemon's choice.

## Test plan

- [x] \`npm run typecheck\` clean
- [x] \`npm run lint\` clean
- [x] \`npm run format:check\` clean
- [x] \`npm test\` — all 141 tests pass
- [x] \`npm run build\` succeeds
- [ ] End-to-end: Go daemon from dicode-core#199 handshakes with this relay build successfully (run manually during review)

## Files

- \`proto/relay.proto\` (vendored from dicode-core)
- \`proto/buf.yaml\`, \`buf.gen.yaml\` — codegen config
- \`src/relay/pb/relay_pb.ts\` — generated (committed)
- \`src/relay/protocol.ts\` — thin re-exports + ForwardResponse interface + hand-typed OAuthTokenDeliveryPayload
- \`src/relay/server.ts\` — handshake/serve dispatch on oneof.kind.case; fromJson/toJson at wire boundaries
[...]

PR #59, opened 2026-04-25
docs(proto): add npm proto script and align README with reality

dr14-make:
## Summary

- `proto/README.md` (added in #57) tells contributors to run `npm run proto`, but no such script existed in `package.json`. Running the documented command failed.
- Added `"proto": "buf generate && prettier --write src/relay/pb"` so the docs work as written.
- Chaining prettier makes the script idempotent — running it on a clean tree produces no diff. Without it, `protoc-gen-es` emits unformatted code that doesn't match the prettier-formatted output committed in #57.

## Test plan

- [x] `npm run proto` succeeds
- [x] After `npm run proto` on a clean tree, `git status` shows no changes to `src/relay/pb/`
- [ ] CI green (lint / test / typecheck unaffected — package.json is the only file touched)

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "format": "prettier --write src tests",
     "format:check": "prettier --check src tests",
     "typecheck": "tsc --noEmit",
+    "proto": "buf generate && prettier --write src/relay/pb",
     "prepublishOnly": "npm run build"
   },
   "dependencies": {


dr14-make (commented):
## Overview

Pure docs/scripts fix: adds the `proto` npm script that `proto/README.md` (#57) already documented but didn't exist. One-line change.

## What's good

- **Idempotency by design.** Chaining `prettier --write src/relay/pb` after `buf generate` is the right call — `protoc-gen-es` emits unformatted code, so without prettier the regen produces a noisy diff against the (formatted) committed output. The commit message captures this rationale clearly.
- **Tight scope.** Touches only the failing line, no opportunistic refactor.

## Suggestions (none blocking)

- **Implicit `buf` prerequisite.** The core repo's `make proto` guards with `command -v buf >/dev/null 2>&1 || { echo "buf not found — run 'make proto-tools' first"; exit 1; }`. This script just fails with a generic `buf: command not found`. Two ways to address (pick one or skip):
  - Add a one-liner to `proto/README.md` like `# requires \`buf\` on PATH (\`brew install bufbuild/buf/buf\` or \`go install github.com/bufbuild/buf/cmd/buf@latest\`)`.
  - Or call `npx --no buf generate` after adding `buf` to devDependencies — keeps the toolchain reproducible.
  - Skip if you're fine with the rough edge for now.
- **CI drift check, eventually.** The `proto/README.md` already has a TODO for upstream-drift CI. A natural sibling: a CI step that runs `npm run proto` and fails if `git diff --exit-code` is non-empty — catches stale regen output. Not for this PR.
- **Script placement (very minor).** Inserted between `typecheck` and `prepublishOnly`; arguably belongs near `build` since it's a build-adjacent codegen step. The existing order isn't strictly alphabetical, so this is taste.

## Risk

None. Pure addition of an npm script entry; doesn't run on install, doesn't affect publish (`prepublishOnly` still calls `npm run build`).

PR #64, opened 2026-05-02
Publish client subpath as dicode-relay/client (v0.2.0)

dr14-make:
## Summary
- Adds a new `dicode-relay/client` subpath export — a TypeScript/Web Crypto client library that owns the WSS handshake, request/response forwarding, TOFU broker pinning, and OAuth helpers (`buildAuthURL`, `decryptTokenEnvelope`).
- Hoists `protocol.ts`, `crypto.ts`, `signing.ts` to `src/shared/` so both the server and the new client consume the same types/helpers (zero behavioral change to the server side).
- Adds `eciesDecryptWebCrypto` (Web Crypto-compatible variant of the existing Node-side `eciesDecrypt`).
- Bumps to `0.2.0` (minor — new exported API).

## Why
This is Part A of dicode-core#253 — migrating the relay client from Go (`pkg/relay/client.go`, ~1,500 LOC + ~1,000 LOC of generated protobuf) to TypeScript. dicode-core's buildin tasks (`relay-client`, `auth-start`, `auth-relay`) will consume this library so they each become ~15-20 LOC of glue.

## Architecture
- `Identity` — Web Crypto P-256 split sign/decrypt keypair, persisted via a `KvAdapter` interface; UUID derived as `hex(sha256(uncompressed_sign_pubkey))`.
- `TofuStore` — broker pubkey pinning (`new` / `match` / `mismatch`).
- `RelayClient` — connection supervisor: dial → handshake → serve → reconnect with exponential backoff (1s→60s, ±20% jitter, reset after 10s stable).
- `buildAuthURL` / `decryptTokenEnvelope` — OAuth flow primitives.
- All wire format JSON via `@bufbuild/protobuf`'s `fromJson`/`toJson` (broker-compatible).

## Test plan
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run test:coverage` — 175 tests, ≥90% line/function coverage
- [x] `npm run build` — produces `dist/client/index.{js,d.ts}` and per-module files
- [x] In-process integration test (`tests/client/client.test.ts`) — boots a real `RelayServer`, connects a real `RelayClient`, asserts a `forward()` round-trips through a real local HTTP listener
- [x] Cross-stack crypto tests:
  - Web Crypto sign → Node `verifyECDSA` (challenge + auth-payload)
  - Node `eciesEncrypt` → Web Crypto `eciesDecryptWebCrypto`
  - Broker `loadBrokerSigningKey` → `verifyDeliverySignature`

## Follow-up
- Part B–D in dicode-core consume this library and delete the Go relay client + protobuf scaffolding.


dr14-make:
## Code Review — dicode-relay PR #64 (relay-client-pkg, v0.2.0)

Reviewed all new `src/client/` files, `src/shared/crypto.ts`, and the test suite. Overall the architecture is solid and the cross-stack crypto wiring is correct. Three issues need addressing before merge, plus a handful of important observations.

---

### Critical

**1. `serve()` leaks the `"message"` listener on abort (`src/client/client.ts:156–175`)**

When `signal` is aborted, `onAbort` calls `ws.terminate()`. That triggers the `"close"` event, which calls `resolve()` — good. But the `"message"` listener registered at line 171 is never removed. If `ws.terminate()` somehow races with an in-flight message, or if the same `ws` instance is reused in a test, the listener leaks. The `"close"` handler should also remove the message listener:

\`\`\`ts
ws.on("close", () => {
  signal?.removeEventListener("abort", onAbort);
  ws.removeListener("message", messageHandler); // add this
  resolve();
});
\`\`\`

**2. `adaptWs` inbox is silently drained on `detach()` — lost frames are unlogged (`src/client/client.ts:234–241`)**

The comment says "discard them; the server shouldn't send extra frames before we've registered our serve listener." This is a TOCTOU window: a fast server could send a `request` frame during the microtask gap between `detach()` (line 134) and `serve()` registering its `"message"` listener. Those frames are permanently lost with no log entry. Either (a) pass the inbox to `serve()` so it can drain any buffered frames, or (b) log a warning if `inbox.length > 0` at detach time so the window is at least observable.

**3. `eciesDecryptWebCrypto` lives in `src/shared/crypto.ts` but uses `node:crypto` Node-only APIs (`createDecipheriv`, `hkdf`) alongside Web Crypto (`crypto.subtle`) — the file ships in the `dicode-relay/client` subpath export that's supposed to be environment-agnostic**

The `shared/crypto.ts` module is imported by both the broker server (Node.js) and the client library. But `eciesDecryptWebCrypto` itself mixes Web Crypto key derivation with `createDecipheriv` from `node:crypto`, which makes it impossible to run in a browser or Cloudflare Worker. This is fine for the current dicode-core consumer (also Node.js), but the CLAUDE.md spec says "environment-agnostic — pass any KvAdapter". Either (a) add a note to `kv-adapter.ts` / README clarifying Node-only, or (b) finish the Web Crypto path (use `SubtleCrypto.decrypt` with AES-GCM instead of `createDecipheriv`) to honour the stated goal.

---

### Important

**4. `signAuthPayload` comment is misleading (`src/client/identity.ts:107–111`)**

The comment says `payloadBuf` is "the 32-byte digest from `buildSignedPayload`". Looking at `src/shared/crypto.ts`, `buildSignedPayload` returns the `.digest()` of a SHA-256 chain — i.e. a 32-byte hash. Web Crypto then signs `sha256(payloadBuf)`, making this a double-hash: `sha256(sha256(...))`. The broker's `createVerify("SHA256").update(payload)` also does one hash pass, so the net result is the same, but the comment should say "we pass the 32-byte pre-computed digest; Web Crypto signs sha256(digest) = sha256(sha256(raw_payload)), broker verifies the same" to prevent future divergence.

**5. `forwarder.ts` does not set a `signal` / timeout on the `fetch` call (line 49)**

A slow or hung local process will hold the relay's pending-request slot open until the relay server's 30 s timeout fires — at which point the relay sends 504 upstream but the client fetch is still running. Add `AbortSignal.timeout(25_000)` (slightly under the relay's 30 s) to `fetchInit` so the client cleans up proactively:

\`\`\`ts
const fetchInit: RequestInit = { method: req.method, headers, signal: AbortSignal.timeout(25_000) };
\`\`\`

**6. `serve()` does not remove the `"error"` listener after `"close"` fires, and vice versa**

The two listeners are never torn down except on their own event. In practice this is harmless on a `ws` WebSocket (the emitter is GCed), but it will produce spurious warnings in test environments that check for listener leaks. Add targeted removes in the `"close"` and `"error"` handlers.

**7. `backoff.ts:next()` jitter ceiling comment is missing**

```ts
return Math.round(Math.min(v * jitter, MAX_MS * (1 + JITTER)));
```

The test documents the 72 s ceiling (`expect(big).toBeLessThanOrEqual(72_000)`) but the source has no comment. A one-liner like `// cap at 72 s (MAX_MS + max jitter)` would make the intent obvious.

---

### Minor / Suggestions

**8. `package.json` is missing `"main"` fallback for CJS consumers**

Dropping `"main"` in favour of `"exports"` is correct for pure ESM, but tools that don't understand `"exports"` (bundlers before Webpack 5, some older Electron versions) will fail to resolve `dicode-relay`. If the consumer base is guaranteed-ESM this is fine — just call it out in the README.

**9. `Identity.import(stored)` is `public` but `StoredIdentity` is not exported (`src/client/identity.ts:49`)**

`Identity.import` accepts a `StoredIdentity` argument, but the type is not re-exported from the barrel. Callers who want to round-trip an `Identity` must construct the object blindly. Either export `StoredIdentity` from `index.ts`, or remove `import` (it appears unused in the public API).

**10. `README.md` typo: "buildin tasks" → "builtin tasks"**

**11. `handshake.ts:69` — log when `welcome.brokerPubkey` is absent**

If the server omits `brokerPubkey` (protocol-valid), the client silently skips TOFU and returns `brokerPubkey: ""`. Callers of `decryptTokenEnvelope` will then pass an empty string, and `verifyDeliverySignature` returns `false` — delivery fails loudly. The failure path is safe, but a `log.warn` at detach time would make the operator-visible signal much clearer.

---

### Strengths

- The `adaptWs` pre-registration pattern (registering the message listener before awaiting "open") is exactly right — the comment explains the microtask-gap hazard clearly.
- DER ↔ P1363 conversion in `identity.ts` is correct and well-tested; the cross-stack test in `tests/client/identity.test.ts` (Node `createVerify` against Web Crypto signature) gives real confidence.
- `dispatchRequest` path allow-list is a good blast-radius limiter.
- The `eciesDecryptWebCrypto` test suite covers session-ID and messageType mismatches — both are important AAD binding checks.
- 175 tests, 90.4% line coverage, and a real in-process integration test (`client.test.ts`) are all well above minimum bar.

---

### Verdict

**Request Changes** — issues 1 (listener leak in `serve()`), 2 (silent frame loss at handshake→serve transition), and 3 (environment-agnostic claim vs Node-only APIs in `eciesDecryptWebCrypto`) should be resolved before merge. Issues 4–6 are important enough to address in the same PR pass.

dr14-make (commented):
## Security Review

Reviewer: Security Review subagent
Scope: New `src/client/` library (identity, auth, forwarder, handshake, tofu) and `src/shared/` refactor introduced in this PR. Only newly-introduced security issues are reported; issues already flagged by the earlier code-review pass are excluded.

---

### CRITICAL — Path Traversal in Request Forwarder

**File:** `src/client/forwarder.ts`, line 29 + 49

**Description:**
The path allow-list check (`startsWith("/hooks/")`) is evaluated on the raw, un-normalized path string. The `fetch` call immediately after constructs a URL by string-concatenating `req.path` into `http://localhost:<port><path>`. The WHATWG URL parser that Node.js `fetch` (undici) uses **normalises `..` segments before sending the request**:

```
/hooks/../admin  →  /admin     (bypasses allow-list)
/hooks/%2e%2e/x →  /x         (bypasses allow-list — percent-decoded before normalisation)
```

**Proof:**
```js
new URL("http://localhost:8080/hooks/../admin").pathname
// → "/admin"
```

**Exploit scenario:**
A compromised or MITM'd relay server sends a `request` frame with `path: "/hooks/../config"` (or any other local endpoint). The forwarder's `startsWith("/hooks/")` check passes, but the fetch target resolves to `http://localhost:<port>/config`, giving the relay read/write access to arbitrary local daemon endpoints — including any unauthenticated administrative APIs.

**Fix:** Normalise the path _before_ the allow-list check, and reject if normalisation changes it:

```typescript
import { posix } from "node:path";

const normalised = posix.normalize(req.path);
if (normalised !== req.path) return errorResponse(req.id, 400);
if (!normalised.startsWith("/hooks/") && normalised !== "/dicode.js") {
  return errorResponse(req.id, 403);
}
```

Alternatively, use `new URL(req.path, "http://localhost").pathname` to parse first and compare the normalised pathname against the allowlist.

---

### IMPORTANT — Inbound `Authorization` and `Cookie` Headers Forwarded to Local Service Without Stripping

**File:** `src/client/forwarder.ts`, lines 38–42

**Description:**
All inbound request headers from the relay server (including `Authorization` and `Cookie`) are forwarded verbatim to the local daemon HTTP service. The `HOP_BY_HOP` set strips connection-management and `Set-Cookie` headers but does **not** strip credential-bearing headers (`Authorization`, `Cookie`, `X-Api-Key`, etc.).

```typescript
for (const [k, hv] of Object.entries(req.headers)) {
  if (k.toLowerCase() === "x-relay-base") continue;  // only x-relay-base is dropped
  for (const v of hv.values) headers.append(k, v);
}
```

**Exploit scenario:**
A compromised relay server (or an attacker who has MITM'd the WSS connection before TOFU is established) injects `Authorization: Bearer <stolen-token>` or a forged session `Cookie` header into a forwarded `request` frame. The local daemon's HTTP handler receives these injected credentials and may grant elevated access.

While the relay is a trusted component, defense-in-depth at the relay→local boundary is warranted: the local service should not receive relay-injected credentials it would treat as user-supplied.

**Fix:** Strip credential headers before forwarding:

```typescript
const STRIP_INBOUND = new Set(["authorization", "cookie", "x-api-key"]);

for (const [k, hv] of Object.entries(req.headers)) {
  const lower = k.toLowerCase();
  if (lower === "x-relay-base") continue;
  if (STRIP_INBOUND.has(lower)) continue;   // add this
  for (const v of hv.values) headers.append(k, v);
}
```

---

### Observations (no action required)

The following were examined and found to be correct:

- **Double-hash in `signAuthPayload`**: `buildSignedPayload` returns a SHA-256 digest; `crypto.subtle.sign({hash:"SHA-256"}, ...)` hashes once more. The broker's `verifyECDSA` → `createVerify("SHA256").update(payload)` applies the same single SHA-256 internally. Both sides consistently sign/verify over `sha256(sha256(...))` — unconventional but internally symmetric. No vulnerability.
- **TOFU mismatch throws**: Confirmed — `handshake.ts:71–75` throws on `"mismatch"`, does not silently continue.
- **`broker_sig` missing check**: `auth.ts:172–174` explicitly throws on absent `broker_sig` before attempting signature verification. Both rejection paths (missing + invalid) are covered and tested.
- **ECDSA sig format (P1363 ↔ DER)**: `p1363ToDer` / `derToP1363` in `identity.ts` correctly handles leading-zero stripping and high-bit padding. DER SEQUENCE length is bounds-checked (rejects ≥ 0x80 multi-byte length). No malleability issue.
- **ECIES auth-tag handling**: Ciphertext is `ct || tag` (last 16 bytes); both `eciesDecrypt` and `eciesDecryptWebCrypto` correctly split at `length - 16`. AES-GCM auth tag is validated before plaintext is returned (Node throws on tag mismatch during `.final()`).
- **HKDF salt binding**: `sessionId` is used as HKDF salt in both encrypt and decrypt paths — ties the derived AES key to the specific OAuth session. AAD (`messageType`) prevents cross-type ciphertext reuse.

dr14-make:
Review iteration #1 complete:

**`/review` findings → commit `1a2e240`:**
- ✅ Critical 1: `serve()` listener leak — refactored to named handlers + shared cleanup
- ✅ Critical 2: handshake→serve frame loss — `detach()` now returns leftover inbox; `runOnce` replays buffered frames before `serve()` listener registers
- ✅ Critical 3: README "environment-agnostic" overstated — replaced with accurate Node 22+/Deno scoping
- ✅ Important 4: misleading double-hash comment in `signAuthPayload` — replaced with explicit explanation
- ✅ Important 5: no fetch timeout in forwarder — `AbortSignal.timeout(25_000)`
- ✅ Important 6: covered by Critical 1 cleanup pattern
- ✅ Minor 7: `StoredIdentity` interface now exported

**`/security-review` findings → commit `6790c10`:**
- ✅ Critical: path traversal bypass — added `decodeURIComponent` + `posix.normalize` guard with 4 adversarial tests (`../`, `%2e%2e`, malformed `%ZZ`, `//`)
- ✅ Important: credential header passthrough — `STRIP_INBOUND` set drops `Authorization`, `Cookie`, `X-Api-Key` from forwarded requests; test confirms

**Final state:** 180 tests (up from 175), coverage 90.39% lines / 92.25% functions, all CI gates green.

Ready for re-review. CC reviewers — please confirm the fixes address your concerns.

dr14-make (commented):
## Re-review: API simplification (commit 88be23e)

Reviewing only what changed in `88be23e` (`refactor(client): remove kv abstraction, make library storage-agnostic`). Earlier findings from iterations 1–2 are excluded from scope.

---

### Confidence in earlier fixes — all preserved ✓

| Earlier fix | Still in place? |
|---|---|
| Listener cleanup in `serve()` | Yes — `cleanup()` removes all three listeners (`close`, `error`, `message`) plus the abort handler (client.ts:185–190). |
| Frame-replay at handshake→serve transition | Yes — `detach()` + replay loop before `serve()` is intact (client.ts:133–147). |
| Path-traversal guard + `STRIP_INBOUND` | Yes — forwarder.ts unchanged; `posix.normalize` guard at line 45, `STRIP_INBOUND` set at line 23, applied at line 61. |
| `signAuthPayload` clarifying comment | Yes — the double-SHA256 explanation is preserved in identity.ts:114–119. |
| Fetch timeout | Yes — `AbortSignal.timeout(25_000)` at forwarder.ts:73. |

None of the earlier fixes were touched by this refactor.

---

### New findings

#### Important — Missing `"match"` TOFU branch in handshake unit tests

**Confidence: High**

`tests/client/handshake.test.ts` tests `"new"` (lines 58, 110, 123) and `"mismatch"` (line 97), but has no test for the `"match"` path — i.e. a second connection where the broker pubkey equals the pinned value and the handshake should succeed. The integration test in `client.test.ts` (line 63) does exercise a reconnect that would hit `"match"` in practice, but there is no isolated unit test asserting that `runHandshake` completes successfully when `tofuCheckAndPin` returns `"match"`. Given that `"match"` is one of the three documented return values in the callback contract, it should have an explicit unit test alongside the other two.

Suggested addition to `handshake.test.ts`:
```ts
it("accepts on TOFU match", async () => {
  const ws = new MockSocket();
  const id = await Identity.generate();
  const tofuCheckAndPin = (_: string): Promise<TofuResult> => Promise.resolve("match");

  const p = runHandshake(ws, id, tofuCheckAndPin);
  ws.inject(challengeJson("00".repeat(32)));
  await Promise.resolve();
  ws.inject(welcomeJson({ url: "wss://x/u/abc/hooks/", brokerPubkey: "KNOWN_PK", protocol: 3 }));

  const result = await p;
  expect(result.brokerPubkey).toBe("KNOWN_PK");
});
```

---

#### Minor — `tofuCheckAndPin` JSDoc does not spell out the "consumer must persist on new" obligation clearly enough

**Confidence: Medium**

The doc comment in `RelayClientOptions` (client.ts:31–33) says:
> "consumer compares against its persisted record, persists if 'new'"

But it reads as an observation rather than a contract. If a consumer returns `"new"` without actually persisting the key, the next connect will return `"new"` again indefinitely — silently accepting a MITM key rotation. A single sentence making this a hard requirement would prevent the misuse:

```ts
/** Trust-on-first-use callback. Receives the broker's announced pubkey (base64 SPKI DER).
 *  MUST compare against the consumer's persisted record and:
 *  - Return "new" after persisting the key (first-seen).
 *  - Return "match" if the key equals the persisted record (normal reconnect).
 *  - Return "mismatch" if the key differs (possible MITM) — RelayClient will throw.
 *  The consumer is responsible for actually writing the key on "new"; failing to
 *  do so causes every connect to be treated as first-seen, defeating TOFU. */
```

Same improvement applies to `runHandshake`'s third parameter, which currently has no JSDoc at all.

---

#### Minor — `Identity.generate()` JSDoc omits that the uuid is derived from the sign key

**Confidence: Low**

The current doc:
> "Generate a fresh split P-256 identity. The caller is responsible for persisting the result of `export()` if they want a stable UUID across runs."

The uuid derivation rule (`hex(sha256(uncompressed_sign_pubkey))`) is important for consumers who need to know their UUID before the first successful handshake (e.g. to build the `hook_base_url`). Worth adding one sentence: `// UUID is hex(sha256(uncompressed sign public key)).`

---

### API surface quality — otherwise clean ✓

- `tofuCheckAndPin` callback shape is clear and composable. Consumers can wire any storage without importing relay internals.
- `Identity.export()` → `Identity.import()` round-trip is symmetric: both sides use `StoredIdentity` with `sign_priv_pkcs8_b64` / `decrypt_priv_pkcs8_b64`. The round-trip test (identity.test.ts:14–21) confirms uuid, signPubkeyB64, and decryptPubkeyB64 are all preserved.
- `StoredIdentity` is exported as `export type` (index.ts:5, identity.ts:6 uses `interface` not `class`), so it's a pure data shape — correct.
- Barrel is minimal: `RelayClient`, `Identity`, `buildAuthURL`, `decryptTokenEnvelope` as values; `RelayClientOptions`, `RelayClientLogger`, `RelayStatus`, `StoredIdentity`, `TofuResult`, `HandshakeResult`, `BuildAuthURLOpts`, `BuildAuthURLResult` as types. `SocketLike` is intentionally not re-exported (good — it's an internal wiring interface).

### README example — correct ✓

- `myKv.get<StoredIdentity>("identity")` returns `StoredIdentity | null`, and the ternary correctly branches on `stored`. The `import` / `generate` + `export` + `set` path is idiomatic.
- The IIFE pattern for "load-or-generate" is a bit verbose but compiles correctly under strict TypeScript (no implicit `any`, no `as`-cast needed).
- `"new" as const` and `"match" as const` are necessary because the return type infers as `string` without them — good that the example includes them, as consumers will need the same pattern.

### Test coverage of new shapes ✓

- `Identity.generate()` is tested for uuid format, pubkey presence (identity.test.ts:7–11).
- Export → import round-trip is explicitly tested (identity.test.ts:14–21) including `decryptPubkeyB64` — better than the old `loadOrGenerate` test which didn't assert on `decryptPubkeyB64`.
- All handshake tests use inline callbacks; `"new"` and `"mismatch"` are directly exercised. Only `"match"` unit test is missing (flagged above as Important).

---

### Verdict: Comment (not blocking, one Important + two Minor)

The refactor is directionally correct and well-executed. The only substantive gap is the missing `"match"` handshake unit test. The two Minor items are doc improvements. None of the earlier fixes regressed.

🤖 Re-review by Claude Code

dr14-make (commented):
## Security re-review: API simplification (commit 88be23e)

**Verdict: APPROVE** — no new security regressions introduced; both prior fixes confirmed preserved.

---

### Earlier fixes confirmed preserved

**[CONFIRMED] Path-traversal fix (`forwarder.ts`)** — `src/client/forwarder.ts` was not touched by `88be23e`. The `decodeURIComponent` + `posix.normalize` double-check guard (lines 37–47) and the `STRIP_INBOUND` set (`"authorization"`, `"cookie"`, `"x-api-key"`) at line 23 are unchanged and fully intact.

**[CONFIRMED] Credential header strip preserved** — `STRIP_INBOUND` still covers all three headers; `forwarder.ts` was not modified in this diff at all.

---

### New API surface review (commit 88be23e)

#### 1. `Identity.export()` correctness — PASS

- `export()` calls `crypto.subtle.exportKey("pkcs8", this.signKey.privateKey)` and `crypto.subtle.exportKey("pkcs8", this.decryptKey.privateKey)` (identity.ts lines 53–55). Both keys are the correct ones.
- Both keys are generated with `extractable: true` (the second positional argument to `crypto.subtle.generateKey`, lines 29 and 33). `fromStored()` also imports them with `extractable: true` (lines 65 and 72). Export will not throw.
- `export()` returns the struct and does nothing else — no `console.log`, no error-message interpolation, no side effects. Clean.

#### 2. TOFU enforcement integrity — PASS

- `runHandshake` still `await`s the callback (`const res = await tofuCheckAndPin(...)`, line 71). Async TOFU implementations cannot be raced due to the await.
- `runHandshake` still throws on `"mismatch"` (lines 72–75). The error message is unchanged.
- The guard is still inside `if (welcome.brokerPubkey)` (line 70), so an absent/empty `brokerPubkey` still skips TOFU silently — correct behaviour for a broker that hasn't announced its key.
- Test coverage: `handshake.test.ts` has an explicit `"rejects on TOFU mismatch"` test where the callback returns `"mismatch"` and the promise is expected to reject. The TOFU path is covered.

#### 3. `signChallenge`/`signAuthPayload` correctness — PASS

Neither `signChallenge` nor `signAuthPayload` was touched by the refactor. The P-256 P1363→DER conversion path is unchanged. The double-hash comment (`sha256(sha256(fields))`) that aligns with Go's `verifyECDSA` path is still present and accurate.

#### 4. `decryptTokenEnvelope` preserved — PASS

`src/client/auth.ts` was not modified. The function still rejects on missing `broker_sig` (line 57) and on a failed `verifyDeliverySignature` (line 69). Both guards are in place.

#### 5. No accidental private-key exposure in tests — PASS

No test in `tests/client/` accesses `signKey`, `decryptKey`, or `sign_priv_pkcs8_b64`/`decrypt_priv_pkcs8_b64` directly. The `identity.test.ts` round-trip test (`export → import`) asserts only on `uuid`, `signPubkeyB64`, and `decryptPubkeyB64` — the public halves. No console logging of key material is present.

#### 6. Storage trust boundary documentation — MINOR / INFO

The README's "Client library" section (lines 266–305) correctly notes that "consumers own all persistence" and shows `id.export()` being persisted to `myKv`. However, there is **no explicit callout** that `StoredIdentity` contains private key material and must be protected like a secret (e.g. encrypted at rest, access-controlled). The Security Model section (lines 206–214) does not mention the consumer-side storage trust boundary.

This is a documentation gap rather than a code defect — the design is sound — but a future integrator reading only the example could plausibly store `StoredIdentity` in a world-readable config file without realising the consequence.

**Recommendation:** Add a one-sentence warning in the README example (e.g. a comment on the `myKv.set("identity", …)` line: `// StoredIdentity contains private key material — treat with the same care as a TLS private key.`) and/or a bullet in the Security Model section. No code change required; this is advisory.

---

### Summary table

| Check | Result | Confidence |
|---|---|---|
| Path-traversal fix preserved (`forwarder.ts`) | CONFIRMED | High — file unmodified |
| `STRIP_INBOUND` credential strip preserved | CONFIRMED | High — file unmodified |
| `Identity.export()` uses pkcs8 for both keys | PASS | High |
| Keys are `extractable: true` at construction | PASS | High |
| `export()` no console/log leakage | PASS | High |
| TOFU callback `await`ed (no race) | PASS | High |
| TOFU throws on `"mismatch"` | PASS | High |
| Empty `brokerPubkey` skips TOFU silently | PASS | High |
| `signChallenge`/`signAuthPayload` unchanged | PASS | High — unmodified |
| `decryptTokenEnvelope` broker_sig + verify | PASS | High — unmodified |
| No private key exposure in new tests | PASS | High |
| README warns `StoredIdentity` is private-key material | **MINOR GAP** | Medium |

No Critical or Important findings. One Minor / advisory documentation gap on the storage trust boundary.

PR #66, opened 2026-05-04
feat: GET /providers endpoint

dr14-make:
## Summary

- Adds public unauthenticated `GET /providers` endpoint that returns a JSON array of configured OAuth providers with metadata (`key`, `pkce`, `scopes`, `secret_required`, `configured`)
- No `clientId` or `clientSecret` values are ever included in the response
- Adds `PublicProviderInfo` interface to `src/broker/providers.ts` (next to `ProviderConfig` for discoverability)
- Full test coverage in `tests/broker/providers-endpoint.test.ts` with explicit `not.toContain` assertions verifying zero secret material leaks via the response body

## Test plan

- [x] `npm run typecheck` — passes
- [x] `npm run lint` — passes
- [x] `npm run format:check` — passes
- [x] `npm run test` — 180/180 tests pass (including 2 new tests for the endpoint)
- [x] `npm run build` — passes
- [x] Test: stripe provider with empty `client_id` is skipped (not in response)
- [x] Test: PKCE-only provider (slack) shows `secret_required: false`
- [x] Test: empty provider map → empty array response
- [x] Test: no `clientId`, `clientSecret`, `client_id`, or `client_secret` keys in any response object
- [x] Test: no secret string values appear anywhere in the serialized response

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
+- `GET /providers` endpoint — returns the list of configured OAuth providers
+  (key, scopes, PKCE flag, secret-required flag, configured flag). No secret
+  material exposed. Consumed by dicode-core's `auth-providers` task.
+
 ## [0.1.4](https://github.com/dicode-ayo/dicode-relay/compare/v0.1.3...v0.1.4) (2026-05-03)
 
 
--- a/src/broker/providers.ts
+++ b/src/broker/providers.ts
@@ -7,6 +7,21 @@
 
 import type { RelayConfig } from "../config.js";
 
+/** Public metadata about a configured OAuth provider — no secret values exposed. */
+export interface PublicProviderInfo {
+  /** Provider key (matches Grant's provider key, e.g. "github", "slack") */
+  key: string;
+  /** Whether PKCE is used */
+  pkce: boolean;
+  /** Default scopes the broker uses if the auth-start task doesn't override */
+  scopes: string[];
+  /** Whether this provider requires a client_secret (true = secret_required) */
+  secret_required: boolean;
+  /** Whether the broker has client_id configured. False means the provider
+   *  is in the YAML but lacks credentials → not enabled in this deployment. */
+  configured: boolean;
+}
+
 /** Runtime provider config with resolved (non-env-ref) values. */
 export interface ProviderConfig {
   /** Grant's provider key — same as the YAML key (e.g. "github", "slack") */
--- a/src/index.ts
+++ b/src/index.ts
@@ -21,7 +21,11 @@ import { RelayServer } from "./relay/server.js";
 import { buildGrantMiddleware } from "./broker/grant.js";
 import { buildBrokerRouter } from "./broker/router.js";
 import { MOCK_PROVIDER_KEY, buildE2EMockRouter, isE2EMockEnabled } from "./broker/e2e-mock.js";
-import { buildProviderMap, type ProviderConfig } from "./broker/providers.js";
+import {
+  buildProviderMap,
+  type ProviderConfig,
+  type PublicProviderInfo,
+} from "./broker/providers.js";
 import { SessionStore } from "./broker/sessions.js";
 import { loadBrokerSigningKey } from "./shared/signing.js";
 import { MetricsCollector } from "./status/metrics.js";
@@ -204,6 +208,20 @@ app.get("/api/status", statusAuth(statusPassword), (_req, res) => {
 // Health check
 app.get("/health", (_req, res) => {
   res.json({ ok: true });
+});
+
+// Public provider metadata — no secret values exposed.
+app.get("/providers", (_req, res) => {
+  const list = Array.from(realProviders.entries()).map(
+    ([key, cfg]): PublicProviderInfo => ({
+      key,
+      pkce: cfg.pkce,
+      scopes: cfg.scopes,
+      secret_required: cfg.clientSecret !== undefined,
+      configured: cfg.clientId !== "",
+    }),
+  );
+  res.json(list);
 });
 
 // Start listening
--- a/src/index.ts
+++ b/src/index.ts
@@ -218,6 +218,10 @@ app.get("/providers", (_req, res) => {
       pkce: cfg.pkce,
       scopes: cfg.scopes,
       secret_required: cfg.clientSecret !== undefined,
+      // buildProviderMap already skips providers with empty clientId, so every
+      // entry in realProviders has a non-empty clientId → always true here.
+      // The field exists for API forward-compatibility with a future
+      // "show unconfigured providers" mode.
       configured: cfg.clientId !== "",
     }),
   );
--- a/tests/broker/providers-endpoint.test.ts
+++ b/tests/broker/providers-endpoint.test.ts
@@ -1,0 +1,150 @@
+/**
+ * GET /providers endpoint tests — verifies metadata exposure and zero secret leakage.
+ */
+
+import { describe, expect, it, beforeAll, afterAll } from "vitest";
+import express from "express";
+import type { Server } from "node:http";
+import type { AddressInfo } from "node:net";
+import { buildProviderMap } from "../../src/broker/providers.js";
+import type { PublicProviderInfo } from "../../src/broker/providers.js";
+import { defaultConfig } from "../../src/config.js";
+
+describe("GET /providers", () => {
+  let server: Server;
+  let port: number;
+
+  beforeAll(async () => {
+    const config = defaultConfig();
+    // Override broker.providers with test fixtures.
+    config.broker.providers = {
+      github: {
+        client_id: "test-github-id",
+        client_secret: "test-github-secret",
+        pkce: true,
+        scopes: ["user", "repo"],
+      },
+      slack: {
+        client_id: "test-slack-id",
+        pkce: true,
+        scopes: ["channels:read"],
+      },
+      // Empty client_id → skipped by buildProviderMap.
+      stripe: {
+        client_id: "",
+        pkce: false,
+        scopes: ["read_write"],
+      },
+    };
+
+    const providerMap = buildProviderMap(config);
+
+    const app = express();
+    app.get("/providers", (_req, res) => {
+      const list = Array.from(providerMap.entries()).map(
+        ([key, cfg]): PublicProviderInfo => ({
+          key,
+          pkce: cfg.pkce,
+          scopes: cfg.scopes,
+          secret_required: cfg.clientSecret !== undefined,
+          configured: cfg.clientId !== "",
+        }),
+      );
+      res.json(list);
+    });
+
+    await new Promise<void>((resolve) => {
+      server = app.listen(0, () => {
+        port = (server.address() as AddressInfo).port;
+        resolve();
+      });
+    });
+  });
+
+  afterAll(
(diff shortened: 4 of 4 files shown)


dr14-make:
## Code Review

### Overview
Clean, minimal addition — correct implementation, no existing routes touched, all CI gates pass.

### Issues / Suggestions

**Minor: `configured` is always `true` in the current implementation — needs an inline comment.**

`buildProviderMap` already silently skips providers with empty `client_id`, so every entry in `realProviders` has a non-empty `clientId`. The `configured: cfg.clientId !== ""` expression will always evaluate to `true` for every entry the endpoint can ever see. The field exists for forward-compatibility, but without a comment it will confuse any future reader who looks at the `PublicProviderInfo` JSDoc ("False means the provider is in the YAML but lacks credentials") and wonders why the field is always `true`.

Suggested addition in `src/index.ts`:

\`\`\`ts
// Note: buildProviderMap already skips providers with empty clientId,
// so every entry here has configured=true. The field exists for API
// forward-compatibility with a future \"show unconfigured\" mode.
configured: cfg.clientId !== "",
\`\`\`

**Low priority: assert `Content-Type` header in tests.**

`res.json()` sets `Content-Type: application/json` automatically, but asserting it explicitly makes the API contract visible to future test readers:

\`\`\`ts
expect(res.headers.get("content-type")).toMatch(/application\/json/);
\`\`\`

### Security
- Using `realProviders` (not `brokerProviders`) is correct — the E2E mock provider is intentionally excluded.
- The `not.toContain` secret-leakage assertions are exactly right.
- No auth needed for public provider metadata — consistent with `/health`.

### Verdict
**Approve** pending the inline comment on `configured`. The logic is correct and the omission is cosmetic, but it will save the next reader a head-scratch.

dr14-make:
## Security Review — PASS

Reviewed: `src/broker/providers.ts`, `src/index.ts` (new route), `tests/broker/providers-endpoint.test.ts`.

### Findings

**Secret / credential exposure — PASS**
The mapping uses an explicit allowlist (5 named fields), not `...cfg` spread. `clientId` and `clientSecret` are never referenced in the output object. `secret_required` is derived as a boolean from `cfg.clientSecret !== undefined` — value never included. Tests assert `not.toContain` for all three secret strings across every response item.

**Unauthenticated access — PASS**
Provider names + scopes + PKCE flags are non-sensitive — same level of disclosure as the existing Grant `/connect/:provider` 404 pattern (unconfigured providers already enumerable). Correct to skip auth; consistent with `/health`.

**Injection surface — PASS**
No request parameters feed into the response. `key` comes from config map keys (operator-controlled YAML), `scopes` from config arrays, `pkce`/`secret_required`/`configured` are booleans. Zero injection surface.

**E2E mock provider exclusion — PASS**
Route reads from `realProviders`, not `brokerProviders`. Mock entry only exists in `brokerProviders` — cannot appear in the response even when `DICODE_E2E_MOCK_PROVIDER` is set.

**DoS — PASS**
O(n) synchronous in-memory map iteration, n < 20 in practice. No external I/O.

### Verdict
No security issues. Safe to merge.

dr14-make:
## Post-merge code review — GET /providers (commit 9421616)

**Verdict: approved with 2 minor observations and 1 forward-compat note. No critical or blocking issues.**

---

### Summary

The change is small (~50 LOC handler + interface, ~150 LOC tests). The security-relevant question — does the endpoint leak secret material? — is answered cleanly: the transform is explicit field-by-field into a dedicated `PublicProviderInfo` type, TypeScript strict mode + `exactOptionalPropertyTypes` means the compiler enforces the shape, and the `not.toHaveProperty` assertions in the tests cover both camelCase and snake_case variants of the credential field names.

---

### Findings

#### Minor 1 — `grantKey` is silently omitted from the response (intentional, but undocumented)

`ProviderConfig` has a `grantKey` field that is set to the same value as the map key. The handler correctly omits it (the transform only picks the five `PublicProviderInfo` fields), but there is no comment noting that `grantKey` was considered and intentionally excluded. In practice `grantKey === key` for every entry `buildProviderMap` produces, so the omission loses nothing, but a future reader refactoring the map might wonder whether it's an oversight.

**Location:** `src/index.ts:214-228` and `src/broker/providers.ts:26-38`  
**Suggestion:** Add a brief comment on the `grantKey` field in `ProviderConfig` noting it matches the map key and is intentionally excluded from the public API.

---

#### Minor 2 — `configured: cfg.clientId !== ""` is always `true` in the current implementation, and the self-contradicting dead-code path carries no test

The comment in `src/index.ts:221-224` correctly explains that `buildProviderMap` filters out empty `clientId` entries before they reach `realProviders`, so `cfg.clientId !== ""` can never be `false`. This means:

1. The expression is dead code today.
2. The test (`tests/broker/providers-endpoint.test.ts:83-107`) asserts `configured: true` for both fixtures but has no test for `configured: false`.

If a future "show unconfigured providers" mode is ever added (the stated rationale for keeping the field), both the handler and the tests will need updating — and without a `false` test path the regression surface is blind.

**Location:** `src/index.ts:224` / `tests/broker/providers-endpoint.test.ts`  
**Suggestion (optional):** Either drop `configured` until the feature exists (simplest), or add a second fixture in the test that manually injects an entry with `clientId: ""` into a test map and asserts `configured: false`. Opening a tracking issue would also work.

---

#### Forward-compat note — E2E mock provider is correctly excluded from `/providers`

The handler reads `realProviders`, not `brokerProviders`. When `DICODE_E2E_MOCK_PROVIDER` is set, `brokerProviders` gains a `mock` entry with `clientId: "mock-e2e-not-a-real-credential"`. Because the handler uses `realProviders`, that entry never appears in the response. This is the correct behavior; calling it out explicitly because the two maps coexist in `src/index.ts` and the distinction matters for anyone modifying the handler.

---

### Checklist

| Item | Result |
|---|---|
| Secret fields (`clientId`, `clientSecret`) absent from response | ✅ Confirmed — explicit field-by-field transform |
| `ProviderConfig` spread not used | ✅ No spread; object literal with exactly five named fields |
| `PublicProviderInfo` is not permissive (`Record<string, unknown>` etc.) | ✅ All fields are typed primitives / `string[]` |
| Route does not shadow `/u/:uuid/...` patterns | ✅ No conflict — `/providers` is a top-level static path |
| Tests cover zero-provider case | ✅ Second test spins up an empty map |
| Numeric-secret leak via `toContain` | N/A — `client_secret` is typed `string` in the Zod schema; no numeric secrets possible |
| Unauthenticated access intentional | ✅ Confirmed — data matches what any operator can read from their own `relay.yaml`; this is the intent |
| E2E mock excluded | ✅ `realProviders` path, not `brokerProviders` |

dr14-make:
## Post-merge security review — GET /providers (9421616)

**Verdict: no actionable findings. One low-severity deployment note.**

---

### 1. Secret material leakage — CLEAN

**Critical category — no issue found.**

The handler in `src/index.ts` (lines 214–229) uses an explicit field-by-field struct literal typed as `PublicProviderInfo`:

```ts
([key, cfg]): PublicProviderInfo => ({
  key,
  pkce: cfg.pkce,
  scopes: cfg.scopes,
  secret_required: cfg.clientSecret !== undefined,
  configured: cfg.clientId !== "",
})
```

- `clientId` and `clientSecret` are never referenced on the right-hand side.
- `secret_required` is a boolean derived from `clientSecret !== undefined` — the value is not echoed.
- `configured` evaluates `clientId !== ""` — the value is not echoed.
- The `PublicProviderInfo` interface has exactly five fields (`key`, `pkce`, `scopes`, `secret_required`, `configured`). Because the mapping is explicit (not a spread), adding new sensitive fields to `ProviderConfig` in a future PR cannot silently leak into this response.
- There are no `try/catch` blocks wrapping the handler. The only fallible operation (`Array.from(...).map(...)`) is a pure in-memory transform; if it throws, Express's default error handler emits a generic 500 HTML page, not a JSON-serialised `ProviderConfig`.
- `buildProviderMap` (`src/broker/providers.ts`) strips the `client_secret` field entirely for PKCE-only providers (`clientSecret` is `undefined`, not `""`), so `secret_required: cfg.clientSecret !== undefined` correctly returns `false` for those rather than leaking the absence string.

### 2. Information disclosure / enumeration — CLEAN (note below)

**Medium category — no code defect; deployment-time note.**

The endpoint is intentionally unauthenticated. The threat (private-broker recon) is real but low-impact: the response reveals which OAuth providers a deployment has credentials for, nothing about the credentials themselves. The list of integrations is generally considered public product information, and dicode-core's auth-providers task requires this endpoint to be callable without a session.

**Deployment note (LOW):** Operators running a strictly private broker (internal tooling, not a public SaaS) may prefer to restrict access to `/providers` at the reverse-proxy layer (e.g., `nginx` `allow` / `deny` directives for internal CIDR). The current docs (`docs/scaling.md`, `README.md`) reference nginx/HAProxy configurations but do not mention this option. No code change is warranted; a one-liner in `docs/providers.md` or the self-host section of `README.md` would suffice if the project wants to surface it. Not filing a tracking issue — low priority.

### 3. DoS amplification — CLEAN

**Low category — no issue found.**

Per-request work is `O(N)` where `N` is provider count (≤ ~14 per `docs/providers.md`). No file I/O, no network I/O, no regex. Response payload is a small JSON array. The handler cannot be exploited for amplification. No in-process rate limiting exists, which is acceptable given that the deployment docs recommend a reverse-proxy tier.

### 4. Cross-cutting concerns — CLEAN

- **Content-Type:** `res.json(list)` sets `Content-Type: application/json; charset=utf-8` automatically via Express.
- **Method handling:** Only `GET /providers` is registered. Express returns 404 for POST/PUT/DELETE on this path by default — correct.
- **CORS:** No `Access-Control-Allow-Origin` header is emitted by this endpoint or globally configured in `src/index.ts`. Appropriate for a daemon-consumed API.

### 5. Tests — CLEAN (one edge-case note)

**`tests/broker/providers-endpoint.test.ts`:**

- Property-level assertions use `not.toHaveProperty` for both camelCase and snake_case variants of `clientId` / `clientSecret` — good.
- String-level assertions use `not.toContain` for each fixture value individually — covers the main leak vectors.
- The empty-map path is covered by a second test.

**Edge case (informational, not a defect):**  
The test asserts `not.toContain("test-github-secret")`, `not.toContain("test-github-id")`, `not.toContain("test-slack-id")`. The `configured` field emits `true` (a boolean), never the `clientId` string itself. This is already the correct behaviour and the tests correctly verify the string is absent.  
One gap: there is no assertion that the response contains *only* the five expected keys (e.g. `Object.keys(p)` has length 5). This would catch a hypothetical future regression where an extra field is added without updating `PublicProviderInfo`. Not a current defect — the explicit struct literal enforces this at compile time — but a `expect(Object.keys(github)).toEqual(["key", "pkce", "scopes", "secret_required", "configured"])` assertion would make the contract machine-verifiable at the HTTP level. Suggesting as a hardening item, not a finding.

### 6. Type safety — CLEAN

The struct literal is typed `PublicProviderInfo` at the call site. TypeScript will reject any field not declared in the interface at compile time, and the interface is not a structural supertype of `ProviderConfig`. Future PRs that add sensitive fields to `ProviderConfig` and accidentally spread them into this handler will be caught by `tsc` before merging.

---

**Summary:** The implementation is correct and the secret-leakage threat (the primary concern for this endpoint) is fully addressed. The only item worth a future commit is a short mention in the self-host docs noting that private-broker operators can lock down `/providers` at the proxy layer. No security issues to fix.

PR #73, opened 2026-05-10
fix(broker, config): mkdir signing-key dir + reject empty critical vars

dr14-make:
## Summary
- **`fix(broker)`** — `loadBrokerSigningKey` now auto-generates and persists a fresh P-256 key at the YAML-configured `broker.signing_key_file` path when the file doesn't exist yet, mkdir -p'ing the parent dir first. Unblocks first-start for dicode-core's buildin/relay-server, which lands `broker.signing_key_file: ${DICODE_DATADIR}/relay/broker-signing.key` without pre-creating the `relay/` subdir. Subsequent loads read the persisted key back so the broker pubkey stays stable across restarts (preserves TOFU pins on daemons).
- **`fix(config)`** — `status.password` and `server.base_url` are now `.min(1).optional()` in the Zod schema. Absent → documented fallback (no auth / `http://localhost:<port>`). Present but empty (literal `""` or unresolved `${VAR}` interpolation) → `ZodError` at `loadConfig` time naming the field and the env var to set. Closes a latent footgun where a missing `STATUS_PASSWORD` env var would silently expose `/status` to the public internet. Provider `client_id`/`client_secret` are intentionally left as-is — empty there is the documented "disable this provider" knob.

## Test plan
- [x] `npm run format` clean
- [x] `npm run lint` clean
- [x] `npm run build` clean
- [x] `npm run test` — 200 passed (was 192; +8 new test cases across `tests/broker/signing.test.ts` and `tests/config.test.ts`)
- [x] Updated 2 existing tests (`tests/start.test.ts`, `tests/index.test.ts`) that were asserting the old "throw ENOENT on missing signing-key file" footgun; they now assert the new success path and that dryRun never touches disk.

## Context
Found during the dicode-core buildin/relay-server review (dicode-ayo/dicode-core#286). Both fixes are pre-emptive — neither is exploitable today thanks to upstream guards in dicode-core — but both are latent footguns for anyone running the relay standalone or for future consumers that don't replicate dicode-core's required-env declarations.

The task brief described the signing-key bug as "calls `writeFileSync(signingKeyFile, pem)` when no key exists at the configured path" — the current code actually called `readFileSync` and threw ENOENT in that case (the write path only targeted the legacy `<cwd>/broker-signing-key.pem` fallback). The fix is functionally identical to what the brief intended: when `signingKeyFile` is set and missing, auto-generate to it (mkdir -p'd) instead of erroring.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -74,16 +74,61 @@ const TlsSchema = z.object({
 const ServerSchema = z
   .object({
     port: z.number().int().default(5553),
-    base_url: z.string().default(""),
+    // base_url is optional (absent → fall back to http://localhost:<port>
+    // via the .transform below), but when explicitly set in YAML it must be a
+    // non-empty string. Catches the common footgun of writing
+    // `base_url: ${BASE_URL}` in relay.yaml with `BASE_URL` unset: env
+    // interpolation collapses to "", and the relay would silently advertise
+    // wss://localhost:<port>/u/... in the welcome message and break OAuth
+    // callback URLs in production.
+    //
+    // .nullish() (instead of .optional()) so YAML forms that produce `null`
+    // (`base_url:`, `base_url: ~`, `base_url: null`) are also accepted as
+    // "absent" — js-yaml parses an empty key value to `null`, not undefined,
+    // and .optional() alone would yield a generic "expected string, received
+    // null" error rather than treating it as the documented "omit to default".
+    // The .transform() collapses null → undefined so the outer .transform
+    // keeps working without a second null check.
+    base_url: z
+      .string()
+      .min(
+        1,
+        "server.base_url must be non-empty — set BASE_URL or remove the key to default to http://localhost:<port>",
+      )
+      .nullish()
+      .transform((v) => v ?? undefined),
     tls: TlsSchema.default(() => TlsSchema.parse({})),
   })
   .transform((s) => ({
     ...s,
-    base_url: s.base_url !== "" ? s.base_url : `http://localhost:${String(s.port)}`,
+    base_url:
+      s.base_url !== undefined && s.base_url !== ""
+        ? s.base_url
+        : `http://localhost:${String(s.port)}`,
   }));
 
 const StatusSchema = z.object({
-  password: z.string().default(""),
+  // password is optional (absent → /status returns 404, dashboard is disabled
+  // via statusAuth(undefined)), but when explicitly set in YAML it must be a
+  // non-empty string. This rejects the footgun of writing
+  // `password: ${STATUS_PASSWORD}` with STATUS_PASSWORD unset: env
+  // interpolation collapses to "", which the previous start.ts logic mapped
+  // back to "no password → no auth" and silently exposed the dashboard.
+  // Operators who want to disable the dashboard must omit the key.
+  //
+  // .nullish() so the common YAML forms `password:` (no value), `password: ~`,
+  // and `password: null` — all parsed by js-yaml as `null` — are treated as
+  // "absent" rather than producing a generic Zod "expected string, received
+  // null" error. The .transform() collapses null → undefined so downstream
+  // `statusAuth(statusCfg.password)` keeps receiving `undefined` for "off".
+  password: z
+    .string()
+    .min(
+      1,
+      "status.password must be non-empty — set STATUS_PASSWORD or remove the key to disable /status (returns 404)",
+    )
+    .nullish()
+    .transform((v) => v ?? undefined),
 });
 
 const RelaySchema = z.object({
--- a/src/shared/signing.ts
+++ b/src/shared/signing.ts
@@ -6,6 +6,7 @@
  *   1. BROKER_SIGNING_KEY_FILE env → read PEM from file
  *   2. BROKER_SIGNING_KEY env → inline PEM string
  *   3. broker.signing_key_file from relay.yaml → read PEM from file
+ *      (must exist; does NOT auto-generate — see loadBrokerSigningKey)
  *   4. Auto-generate to <cwd>/broker-signing-key.pem on first start
  *
  * Env takes precedence over YAML so operators can override a baked-in
@@ -25,8 +26,8 @@ import {
   createVerify,
   generateKeyPairSync,
 } from "node:crypto";
-import { existsSync, readFileSync, writeFileSync } from "node:fs";
-import { join } from "node:path";
+import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
+import { dirname, join } from "node:path";
 
 const AUTO_KEY_FILENAME = "broker-signing-key.pem";
 
@@ -40,12 +41,23 @@ export interface BrokerSigningKey {
 /**
  * Load or generate the broker's signing key.
  *
+ * Resolution semantics:
+ *   - If `signingKeyFile` is set (from `broker.signing_key_file`), the file
+ *     MUST exist; this branch never auto-generates. Operators who land an
+ *     explicit path opt into managing the key lifecycle themselves —
+ *     auto-generating at a typo'd path would silently rotate the broker
+ *     pubkey and break every TOFU-pinned daemon (see issue #54).
+ *   - If no env/inline/YAML source is set, the legacy cwd fallback at
+ *     `<cwd>/broker-signing-key.pem` is consulted; on first start it is
+ *     auto-generated (with `mkdir -p` for the parent dir, in case cwd's
+ *     parent doesn't exist yet).
+ *
  * @param env                process.env (or test override)
  * @param cwd                working directory for auto-generated key fallback
  * @param signingKeyFile     `broker.signing_key_file` from relay.yaml (or "" if unset)
  * @param allowAutoGenerate  when no env/file source is present, controls the
- *                           fallback: `true` (default — legacy CLI behavior)
- *                           persists a freshly generated key to
+ *                           legacy cwd-fallback: `true` (default — legacy CLI
+ *                           behavior) persists a freshly generated key to
  *                           `<cwd>/broker-signing-key.pem`; `false` returns an
  *                           in-memory ephemeral key without touching disk.
  *                           `startServer({ dryRun: true })` passes `false` so
@@ -65,8 +77,24 @@ export function loadBrokerSigningKey(
   } else if (env.BROKER_SIGNING_KEY) {
     pem = env.BROKER_SIGNING_KEY;
   } else if (signingKeyFile !== "") {
+    // A YAML-configured `broker.signing_key_file` was provided. This path is
+    // operator-trusted: do NOT auto-generate at it if it's missing — that
+    // would mask typos (rotating the broker pubkey to a wrong location and
+    // breaking TOFU-pinned daemons). Issue #54 tracks a broader hardening of
+    // when auto-gen is allowed to fire at all. Surface ENOENT with a clear
+    // message naming the path so operators can spot the typo or pre-create
+    // the file via their orchestration layer.
+    if (!existsSync(signingKeyFile)) {
+      throw new Error(
+        `broker.signing_key_file points to a missing file: ${signingKeyFile}. ` +
+          `Either create the file (e.g. generate a P-256 PEM with openssl) or ` +
+          `unset broker.signing_key_file to fall back to auto-generation.`,
+      );
+    }
     pem = readFileSync(signingKeyFile, "utf8");
   } else {
+    // No env, no inline key, no YAML path. Fall back to the legacy
+    // cwd-relative location and auto-generate on first start.
     const autoPath = join(cwd, AUTO_KEY_FILENAME);
     if (existsSync(autoPath)) {
       pem = readFileSync(autoPath, "utf8");
@@ -86,9 +114,16 @@ export function loadBrokerSigningKey(
         publicKeyEncoding: { type: "spki", format: "pem" },
         privateKeyEncoding: { type: "pkcs8", format: "pem" },
       });
+      // Safety: cwd may itself be a freshly-created path where the parent
+      // directory doesn't yet exist. mkdir -p before writeFileSync so we
+      // don't throw ENOENT on the legacy fallback. This is the narrow fix
+      // the original brief intended — the YAML-configured path branch
+      // above deliberately does NOT auto-generate (see #54 for context).
+      mkdirSync(dirname(autoPath), { recursive: true });
       writeFileSync(autoPath, pair.privateKey, { mode: 0o600 });
-      console.log(
-        `broker: generated signing key at ${autoPath} — set BROKER_SIGNING_KEY_FILE to use a persistent key`,
+      console.warn(
+        `broker: generated signing key at ${autoPath} — ` +
+          `set BROKER_SIGNING_KEY_FILE or broker.signing_key_file to use a persistent key`,
       );
       pem = pair.privateKey;
     }
--- a/tests/broker/signing.test.ts
+++ b/tests/broker/signing.test.ts
@@ -2,8 +2,9 @@
  * Broker signing key tests — key loading, signing, verification round-trip.
  */
 
-import { writeFileSync, unlinkSync, existsSync } from "node:fs";
-import { join } from "node:path";
+import { mkdtempSync, rmSync, writeFileSync, unlinkSync, existsSync, statSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { dirname, join } from "node:path";
 import { generateKeyPairSync } from "node:crypto";
 import { afterEach, describe, expect, it } from "vitest";
 import {
@@ -74,6 +75,75 @@ describe("loadBrokerSigningKey", () => {
       tmpKeyPath,
     );
     expect(key.publicKeyBase64).toBeTruthy();
+  });
+
+  it("throws ENOENT (does NOT auto-generate) when signing_key_file points at a missing file", () => {
+    // Operator-trusted path: a typo'd `signing_key_file` must surface ENOENT
+    // on first start rather than silently auto-generating at the wrong path
+    // (which would rotate the broker pubkey and break TOFU-pinned daemons —
+    // see issue #54). The legacy cwd-fallback below is the only auto-gen
+    // surface this loader exposes.
+    const baseTmp = mkdtempSync(join(tmpdir(), "dicode-relay-signing-"));
+    const targetPath = join(baseTmp, "non-existent-subdir", "broker-signing.key");
+    try {
+      expect(existsSync(dirname(targetPath))).toBe(false);
+
+      expect(() =>
+        loadBrokerSigningKey(
+          { BROKER_SIGNING_KEY_FILE: "", BROKER_SIGNING_KEY: "" },
+          process.cwd(),
+          targetPath,
+          true,
+        ),
+      ).toThrow(/broker\.signing_key_file points to a missing file/);
+
+      // Importantly: no file or parent dir was created as a side effect.
+      expect(existsSync(targetPath)).toBe(false);
+      expect(existsSync(dirname(targetPath))).toBe(false);
+    } finally {
+      rmSync(baseTmp, { recursive: true, force: true });
+    }
+  });
+
+  it("mkdir -p's the parent directory on the legacy cwd-fallback auto-generate path", () => {
+    // The legacy fallback (no env, no inline, no YAML path) auto-generates a
+    // P-256 key under `<cwd>/broker-signing-key.pem`. If cwd itself is a
+    // freshly-created tmpdir whose parent path doesn't yet exist (or if a
+    // future caller passes a synthetic cwd via the second arg), writeFileSync
+    // would throw ENOENT. mkdir -p the parent so the narrow fix the brief
+    // intended actually fires here.
+    const baseTmp = mkdtempSync(join(tmpdir(), "dicode-relay-cwd-"));
+    const syntheticCwd = join(baseTmp, "fresh-subdir");
+    expect(existsSync(syntheticCwd)).toBe(false);
+    try {
+      const key = loadBrokerSigningKey(
+        { BROKER_SIGNING_KEY_FILE: "", BROKER_SIGNING_KEY: "" },
+        syntheticCwd,
+        "", // no YAML path → legacy fallback branch
+        true,
+      );
+      const autoPath = join(syntheticCwd, "broker-signing-key.pem");
+      expect(existsSync(autoPath)).toBe(true);
+      expect(key.publicKeyBase64).toBeTruthy();
+      const payload = buildDeliverySignaturePayload("t", "s", "e", "c", "n");
+      const sig = key.sign(payload);
+      expect(verifyDeliverySignature(key.publicKeyBase64, sig, "t", "s", "e", "c", "n")).toBe(true);
+      if (process.platform !== "win32") {
+        const mode = statSync(autoPath).mode & 0o777;
+        expect(mode).toBe(0o600);
+      }
+
+      // Second load must read the persisted key back (no rotation).
+      const second = loadBrokerSigningKey(
+        { BROKER_SIGNING_KEY_FILE: "", BROKER_SIGNING_KEY: "" },
+        syntheticCwd,
+        "",
+        true,
+      );
+      expect(second.publicKeyBase64).toBe(key.publicKeyBase64);
+    } finally {
+      rmSync(baseTmp, { recursive: true, force: true });
+    }
   });
 
   it("env BROKER_SIGNING_KEY_FILE takes precedence over YAML signing_key_file", () => {
--- a/tests/start.test.ts
+++ b/tests/start.test.ts
@@ -240,17 +240,18 @@ broker:
     await handle.close();
   });
 
-  it("rejects when signing_key_file points at an unreadable path", async () => {
-    // Use a path under `ctx.dir` that doesn't exist — fails with ENOENT
-    // identically on Linux and macOS (the previous /proc/1/... approach
-    // worked on Linux runners but produced a confusing error on macOS
-    // developer machines that don't have /proc).
-    const unreadable = join(ctx.dir, "nonexistent-dir", "broker-signing.key");
+  it("rejects a signing_key_file pointing at a missing file (configured-path branch never auto-generates)", async () => {
+    // Operator-trusted YAML path: a typo'd `signing_key_file` must surface a
+    // clear error rather than silently auto-generating at the wrong path
+    // (which would rotate the broker pubkey and break TOFU-pinned daemons).
+    // dryRun still exercises the same loader, so the error fires there too.
+    // Issue #54 tracks broader hardening of auto-gen surface area.
+    const target = join(ctx.dir, "nonexistent-dir", "broker-signing.key");
     writeYaml(
       ctx,
       `
 broker:
-  signing_key_file: ${unreadable}
+  signing_key_file: ${target}
 `,
     );
 
@@ -260,7 +261,10 @@ broker:
         env: { ...process.env },
         dryRun: true,
       }),
-    ).rejects.toThrow();
+    ).rejects.toThrow(/broker\.signing_key_file points to a missing file/);
+
+    // Importantly: no file or parent dir was materialised as a side effect.
+    expect(existsSync(target)).toBe(false);
   });
 });
 
--- a/relay.yaml.example
+++ b/relay.yaml.example
@@ -13,8 +13,14 @@ server:
     cert_file: ""          # e.g. ${TLS_CERT_FILE}
     key_file: ""           # e.g. ${TLS_KEY_FILE}
 
-status:
-  password: ${STATUS_PASSWORD}
+# status:
+#   # Uncomment and set STATUS_PASSWORD in the environment to enable the
+#   # password-protected /status dashboard (basic auth, password-only — the
+#   # username is ignored). Leaving the entire section commented out, or
+#   # omitting `password`, makes /status and /api/status return 404
+#   # ("status page not configured"). An empty value is rejected at load
+#   # time so a typo'd STATUS_PASSWORD never silently exposes the dashboard.
+#   password: ${STATUS_PASSWORD}
 
 relay:
   timestamp_tolerance_s: 30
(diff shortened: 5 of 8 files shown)


src/shared/signing.ts
```diff
@@ -86,6 +89,12 @@ export function loadBrokerSigningKey(
         publicKeyEncoding: { type: "spki", format: "pem" },
         privateKeyEncoding: { type: "pkcs8", format: "pem" },
       });
+      // Ensure the parent directory exists. Consumers like dicode-core's
+      // buildin/relay-server land `broker.signing_key_file:
+      // ${DICODE_DATADIR}/relay/broker-signing.key`, but core does not
+      // pre-create the `relay/` subdir — writeFileSync would otherwise
+      // throw ENOENT on first start.
+      mkdirSync(dirname(autoPath), { recursive: true });
```
dr14-make, line 122: warning: this `mkdirSync` widens the auto-generate write-target from the previously-fixed `<cwd>/broker-signing-key.pem` to any path the operator writes in `relay.yaml`. That's a deliberate UX win for first-start in dicode-core's buildin/relay-server template, but it also turns a typo'd `signing_key_file` path into a silent state-bleed: instead of ENOENT on first start, the relay now creates a fresh key at the wrong location, runs with it, and any TOFU-pinned daemon refuses the connection (defense in depth, but the misconfiguration accumulates stale keys across typo'd paths instead of surfacing immediately).

Open question on the relationship to #54 ("gate auto-gen behind NODE_ENV/opt-in"): that issue argues for *narrowing* the auto-gen surface in production; this commit *widens* it. Consider whether the mkdir path should additionally require `env.NODE_ENV !== 'production'` OR an explicit `broker.allow_auto_generate: true` YAML flag, mirroring the `BROKER_ALLOW_EPHEMERAL_KEY` lever proposed in #54. The fleet-DoS scenario from #54 (ephemeral container without persistent volume) still applies whether the auto-path is in cwd or under `${DICODE_DATADIR}/relay/`.

Defense-in-depth note worth a code comment here: `mkdirSync(..., { recursive: true })` follows symlinks on Linux, so a hostile symlinked parent directory could write the private key outside the intended location. Same trust boundary as before (the operator controls `relay.yaml`), but the comment should call out that `signing_key_file` is operator-trusted input — if any consumer ever sources it from user-controllable data, this becomes an arbitrary-file-write primitive.

src/config.ts
```diff
@@ -74,16 +74,44 @@ const TlsSchema = z.object({
...
+  // explicitly set in YAML it must be a non-empty string. This rejects the
+  // footgun of writing `password: ${STATUS_PASSWORD}` with STATUS_PASSWORD
+  // unset: env interpolation collapses to "", which the previous start.ts
+  // logic mapped back to "no password → no auth" and silently exposed the
+  // dashboard. Operators who want a public dashboard must omit the key.
+  password: z
+    .string()
+    .min(
+      1,
+      "status.password must be non-empty — set STATUS_PASSWORD or remove the key to leave /status public",
```
dr14-make: info: the error message and the surrounding comment claim "absent → no auth, public dashboard" / "remove the key to leave /status public". That's not what `src/status/auth.ts` actually does: `statusAuth(undefined)` returns a 404 "status page not configured" handler, *not* an unauthenticated dashboard. The commit message has the same wording.

This is just a docs/error-message accuracy issue — the security posture is fine (404 is strictly safer than "public") — but the inaccurate guidance will confuse an operator who actually wants a public dashboard and follows the error's advice (omitting the key) only to find /status returning 404.

Suggest: `"... or remove the key to disable /status (returns 404)"`. Same fix on the matching commit-message paragraph.

src/config.ts
```diff
@@ -74,16 +74,44 @@ const TlsSchema = z.object({
...
+    )
+    .optional(),
```
dr14-make: info: `.optional()` accepts `undefined` but rejects `null`. YAML lines like `password:` (key with no value), `password: ~`, or `password: null` all parse to `{ password: null }`, which produces a Zod error of the shape `"Invalid input: expected string, received null"` — without the helpful "set STATUS_PASSWORD or remove the key" hint, because that hint is only attached to the `.min(1)` check.

Two options:
1. Add `.nullable()` and treat `null` as "absent" (matches what an operator writing `password:` likely meant — same intent as omitting the key).
2. Leave as-is; verify-by-test that the `null` path produces *some* error (already true), and accept the worse error message.

Either is fine. Same applies to `server.base_url` below. Recommend (1) plus a test case for `password:` (no value) and `password: null` to lock in the expected behaviour either way.

dr14-make:
## Security review

**Verdict: APPROVE with non-blocking warnings.** Both fixes are net-positive for security posture. No blocking findings.

### Confirmed-good

- **Empty `${VAR}` rejection for `status.password` and `server.base_url`**: real fail-closed behaviour. The previous `statusCfg.password !== "" ? password : undefined` path that silently mapped unresolved env to no-auth is now impossible.
- **dryRun preserves the #71 invariant**: `startServer({ dryRun: true })` passes `allowAutoGenerate = false`, which short-circuits in `loadBrokerSigningKey` at line 76 before reaching `mkdirSync`/`writeFileSync`. New `tests/start.test.ts` test asserts `existsSync(target) === false` after dryRun. Verified.
- **Subsequent loads reuse persisted key** (TOFU pubkey stays stable across restarts). Test covers it.
- **Key file mode 0o600** preserved on the new mkdir+write path. Test asserts.
- **No new write surface from untrusted input**: `signing_key_file` is sourced exclusively from `broker.signing_key_file` in operator-rendered YAML. There is no env-var override for the *path* (env overrides only the *content* via `BROKER_SIGNING_KEY_FILE` / `BROKER_SIGNING_KEY`, which take precedence anyway). Trust boundary is unchanged: an operator who can write `relay.yaml` can already write arbitrary files via other means.
- **Test fixtures contain no realistic-shaped credentials**: tests use unresolved `${STATUS_PASSWORD}` references and literal `""`, never strings like `test-pw` / `secret123`. No `console.log` of key material in new tests.

### Findings (non-blocking)

| Severity | Where | Finding |
|---|---|---|
| warning | `src/shared/signing.ts:97` | mkdir widens auto-gen surface; tension with #54; symlink-following on the parent dir. [Inline](https://github.com/dicode-ayo/dicode-relay/pull/73#discussion_r3215530425) |
| info | `src/config.ts:112` | Error message and comment incorrectly say absent password results in "public dashboard"; `statusAuth(undefined)` actually returns 404. [Inline](https://github.com/dicode-ayo/dicode-relay/pull/73#discussion_r3215530719) |
| info | `src/config.ts:114` | `.optional()` rejects YAML `null` (e.g. `password:` with no value) with a generic Zod error, not the helpful hint. Consider `.nullable()` + a test. [Inline](https://github.com/dicode-ayo/dicode-relay/pull/73#discussion_r3215530889) |

### Specific recommendation on the auto-gen widening

**Accept this PR as-is, then open a follow-up issue to address #54 holistically** (not as part of this PR). Two reasons:

1. The widening is genuinely necessary for the dicode-core buildin/relay-server template path (`${DICODE_DATADIR}/relay/broker-signing.key`) to work on first start without a pre-flight `mkdir` script. Reverting it would block dicode-core#286.
2. #54's correct fix is structural — gate auto-gen behind `NODE_ENV !== 'production'` OR `broker.allow_auto_generate: true` (or `BROKER_ALLOW_EPHEMERAL_KEY=1` env), which is a separate design decision from "mkdir the parent dir". Bolting it onto this PR would either over-scope it or land a half-measure. The PR's commit-message framing ("latent footguns for anyone running the relay standalone") explicitly defers the broader hardening — that's the right call.

The two are complementary, not in conflict: this PR makes auto-gen *work correctly* at the YAML-configured path; #54 makes it *refuse to fire* in production unless explicitly opted in. Resolve in that order.

dr14-make (commented):
Overall: well-scoped follow-up; the two bugs are real and the test coverage is mostly tight. A few issues to address before merge — most importantly the `.optional()` vs YAML-`null` mismatch in StatusSchema (the very footgun this PR is trying to close has a still-live variant), and an explicit conflict with the still-open #54 design discussion. Inline comments below.

Summary of findings: 2 blocking, 5 nits.

Key items:
1. **blocking** — `z.string().min(1).optional()` rejects `null`, but `js-yaml` produces `null` for the common YAML form `password:` (empty value, including `~` and `null`). That throws a generic `Invalid input: expected string, received null` ZodError that does not name `STATUS_PASSWORD` — the opposite of the helpful messaging this PR set out to provide. Use `.nullish()` (or pre-coerce null → undefined) and add a test.
2. **blocking / scope** — the auto-generate-at-YAML-path widening directly conflicts with the design discussion in #54, which proposes *narrowing* auto-gen in production. The PR body / commit should call out the conflict; either split this into a separate PR after a design decision on #54, or update #54 with the new context.
3. **nit (DX-blocking, file not in diff)** — `relay.yaml.example` line 17 still ships `password: ${STATUS_PASSWORD}`. After this PR, that line hard-fails `loadConfig` whenever `STATUS_PASSWORD` is unset (the example's default). A fresh `cp relay.yaml.example relay.yaml && dicode-relay` then emits `ZodError: status.password must be non-empty` instead of the documented "no auth" default. Drop or comment the line in this PR; suggested replacement: `# password: ${STATUS_PASSWORD}  # uncomment to enable basic auth`.
4. **nit** — `README.md` doesn't mention the new auto-gen-at-YAML-path behavior. Operators may be surprised that `signing_key_file: /path/that/doesnt/exist` now silently creates a key there instead of erroring.
5. **nit** — `console.log` vs `console.warn` inconsistency in the auto-gen branch; the rest of the file logs operator-facing notices via `console.warn`.

src/config.ts
dr14-make: **blocking:** `.optional()` accepts `undefined` but not `null`. `js-yaml` parses the common YAML form `password:` (key present, value empty) as `null`, and the same applies to `password: ~` and `password: null`. With the current schema, those throw a generic Zod error `Invalid input: expected string, received null` that does **not** include the curated `"status.password must be non-empty — set STATUS_PASSWORD or remove the key…"` message — the exact opposite of this PR's intent.

Reproducer:

```js
const { load } = require('js-yaml');
load('status:\n  password:\n')   // → { status: { password: null } }
load('status:\n  password: ~\n')  // → { status: { password: null } }
```

With the current `.min(1).optional()` schema:

```
undefined  => {}                                  // OK (omitted)
''         => ERROR: must be non-empty            // OK (curated)
null       => ERROR: expected string, received null    // BAD (generic, no STATUS_PASSWORD mention)
'hunter2'  => { password: 'hunter2' }              // OK
```

Switch to `.nullish()` and either accept `null` as "absent" or pre-transform it to `undefined`:

```ts
password: z
  .string()
  .min(1, "status.password must be non-empty — set STATUS_PASSWORD or remove the key to leave /status public")
  .nullish()
  .transform((v) => v ?? undefined),
```

The `.transform` keeps `statusCfg.password === undefined` downstream, so `statusAuth(undefined)` keeps working without further changes.

Add a test:

```ts
it("accepts an explicit-empty status.password (YAML `password:`)", () => {
  writeFileSync(tmpPath, 'status:\n  password:\n');
  const cfg = loadConfig({ RELAY_CONFIG: tmpPath });
  expect(cfg.status.password).toBeUndefined();
});
```

Same fix applies to `server.base_url` below.

src/config.ts
```diff
@@ -74,16 +74,44 @@ const TlsSchema = z.object({
...
+    // interpolation collapses to "", and the relay would silently advertise
+    // wss://localhost:<port>/u/... in the welcome message and break OAuth
+    // callback URLs in production.
+    base_url: z
+      .string()
+      .min(
+        1,
+        "server.base_url must be non-empty — set BASE_URL or remove the key to default to http://localhost:<port>",
+      )
+      .optional(),
```
dr14-make: **nit:** same `.optional()` vs `null` issue as for `status.password` — `base_url:` (empty value) in YAML parses to `null` and hits the generic Zod error path instead of the curated one. Use `.nullish().transform((v) => v ?? undefined)` here too, then the existing `s.base_url !== undefined && s.base_url !== ""` check in the outer `.transform` keeps working as-is.

Separately worth considering (out of scope for this PR, but a follow-up issue): should `base_url` be **required** (`.min(1)` without `.optional()`)? The localhost fallback is harmless for `dicode-relay --check` and local dev, but the welcome message advertises `wss://localhost:5553/u/...` to every connecting daemon. For any non-localhost deployment this is broken. At minimum, log a warning at startup if the resolved welcome URL hostname is `localhost`.

src/shared/signing.ts
```diff
@@ -64,10 +64,13 @@ export function loadBrokerSigningKey(
     pem = readFileSync(env.BROKER_SIGNING_KEY_FILE, "utf8");
   } else if (env.BROKER_SIGNING_KEY) {
     pem = env.BROKER_SIGNING_KEY;
-  } else if (signingKeyFile !== "") {
+  } else if (signingKeyFile !== "" && existsSync(signingKeyFile)) {
```
dr14-make: **blocking / scope:** this widens the auto-gen surface in exactly the direction that issue #54 is trying to *narrow*. #54 wants to gate auto-gen behind `NODE_ENV !== 'production'` or an explicit `BROKER_ALLOW_EPHEMERAL_KEY=1`, because every unintended auto-gen rotates the broker pubkey and breaks TOFU-pinned daemons fleet-wide.

After this PR, a typo'd `signing_key_file` value will silently:
1. `mkdir -p` an unintended directory tree (e.g. typo `${DICODE_DATATIR}` → `""` interpolation under the warning path → relative path under cwd that gets created)
2. Auto-generate a fresh key under a path the operator didn't intend
3. Rotate the broker pubkey on every restart from that wrong path

**For dicode-core consumers** (buildin/relay-server with a known-good DICODE_DATADIR path): this fix is correct and necessary.

**For standalone CLI users**: this is exactly the fleet-wide-DoS scenario #54 is trying to prevent — and worse, the surface is now wider because the auto-gen branch fires not just for unset `signing_key_file` but for any set-but-missing path.

Suggested tightenings (any one would make this safer):
- Require an explicit opt-in env var (`BROKER_ALLOW_EPHEMERAL_KEY=1`) — same one #54 proposes — and have buildin/relay-server in dicode-core set it in the child process env.
- Require the parent dir's *parent* to already exist (single-level `mkdir -p` only; unbounded recursive masks typos like `/srv/relay/data/keys/broker.pem` when there's no `/srv/relay/data/` at all).
- Require the file path to end in a known suffix (`.pem` or `.key`).

If any of those feel too heavy for this PR, please at minimum:
1. Add a comment to #54 linking here, explaining that its proposed gate now needs to apply to **both** the cwd-fallback and the YAML-path branches.
2. Add a paragraph to the PR description acknowledging the scope-widening and its interaction with #54.

Ideally split into two PRs — the `mkdir -p` for the legacy cwd-fallback branch is uncontroversial; the auto-gen-at-YAML-path widening deserves its own discussion with #54 as context.

src/shared/signing.ts
```diff
@@ -86,6 +89,12 @@ export function loadBrokerSigningKey(
...
       });
+      // Ensure the parent directory exists. Consumers like dicode-core's
+      // buildin/relay-server land `broker.signing_key_file:
+      // ${DICODE_DATADIR}/relay/broker-signing.key`, but core does not
+      // pre-create the `relay/` subdir — writeFileSync would otherwise
+      // throw ENOENT on first start.
+      mkdirSync(dirname(autoPath), { recursive: true });
       writeFileSync(autoPath, pair.privateKey, { mode: 0o600 });
       console.log(
         `broker: generated signing key at ${autoPath} — set BROKER_SIGNING_KEY_FILE to use a persistent key`,
```
dr14-make: **nit:** `console.log` here, but operator-facing diagnostics in this codebase use `console.warn` (see `resolveEnvVars` in `src/config.ts`). This message reports a side-effect the operator should pay attention to (a new persistent key was just written, plus a parent directory was just created); `console.warn` routes it to stderr and is consistent with the rest of the loader's diagnostics. Also worth mentioning the `dirname(autoPath)` in the message so operators can spot a typo'd path:

```ts
console.warn(
  `broker: generated signing key at ${autoPath} (created parent dir ${dirname(autoPath)}). ` +
  `Persist this file across restarts to keep broker pubkey stable; ` +
  `set BROKER_SIGNING_KEY_FILE or BROKER_SIGNING_KEY to override.`,
);
```

src/shared/signing.ts
dr14-make, line 122: **nit:** `mkdirSync(..., { recursive: true })` uses the process umask for new directories (typically `0o755`). For a directory that's about to hold a `0o600` private key, consider passing `mode: 0o700`:

```ts
mkdirSync(dirname(autoPath), { recursive: true, mode: 0o700 });
```

Then the existing test can assert the created parent dir is `0o700` and `0o600` on the key file together — both surfaces tightened to the same threat model.

tests/broker/signing.test.ts
```diff
@@ -76,6 +77,67 @@ describe("loadBrokerSigningKey", () => {
...
+      expect(existsSync(dirname(targetPath))).toBe(true);
+      // Returned key is usable
+      expect(key.publicKeyBase64).toBeTruthy();
+      const payload = buildDeliverySignaturePayload("t", "s", "e", "c", "n");
+      const sig = key.sign(payload);
+      expect(verifyDeliverySignature(key.publicKeyBase64, sig, "t", "s", "e", "c", "n")).toBe(true);
+      // Permission tightening: 0o600 (skip on platforms that don't honor it)
+      if (process.platform !== "win32") {
+        const mode = statSync(targetPath).mode & 0o777;
+        expect(mode).toBe(0o600);
```
dr14-make, line 133: **nit:** the test asserts `0o600` on the file but not on the parent directory. If you take the `mkdirSync(..., { mode: 0o700 })` suggestion above, add:

```ts
if (process.platform !== "win32") {
  const parentMode = statSync(dirname(targetPath)).mode & 0o777;
  expect(parentMode).toBe(0o700);
}
```

Note: `mkdtempSync` produces a fresh directory per call, so the test is airtight against leaked state between runs — good.

tests/start.test.ts
```diff
@@ -240,27 +240,29 @@ broker:
...
-        env: { ...process.env },
-        dryRun: true,
-      }),
-    ).rejects.toThrow();
+    const handle = await startServer({
+      configPath: ctx.configPath,
+      env: { ...process.env },
+      dryRun: true,
+    });
+    await handle.close();
```
dr14-make: **nit:** the rewritten test correctly asserts the dryRun → no-disk-writes invariant, but loses coverage of the *real-start* path through `startServer` (the path that actually mkdirs + writes). Consider adding a sibling test that calls `startServer({ dryRun: false, env })` against a path under `ctx.dir` and asserts (a) the file was created with mode `0o600`, (b) the broker pubkey is stable across two starts. The unit test in `tests/broker/signing.test.ts` exercises `loadBrokerSigningKey` in isolation, but doesn't cover the integration through `startServer`.

Not strictly needed if you're confident the dryRun-vs-real branching in `start.ts` is well-tested elsewhere, but the deletion of this test does leave a small coverage gap on the integration.

dr14-make:
Thanks for the careful reviews. Pushed `2beb2ae` (`--force-with-lease`) — final two-commit state, history rewritten.

### What was kept
- **Empty `${VAR}` rejection** for `status.password` and `server.base_url` (the original brief).
- **`mkdir -p` on the legacy cwd-fallback** auto-generate path — the genuine narrow safety patch.
- All inline error messaging that names the offending field + env var.

### What was reverted and why
- **Auto-generate at the YAML-configured `signing_key_file` path.** Both reviews flagged this as widening the auto-gen surface in exactly the direction #54 wants to *narrow*. The configured-path branch now throws ENOENT with a clear message naming the missing path; the legacy cwd-fallback remains the only auto-gen surface. This keeps #54's design space open (gate auto-gen behind `NODE_ENV` or an explicit opt-in env var) and avoids silently rotating the broker pubkey on typo'd paths.

### Inline findings (in review order)
| # | Finding | Disposition |
|---|---|---|
| 1 | `mkdirSync` widens auto-gen + symlink-following | **Fixed** (reverted to legacy-fallback-only) |
| 2 | "absent → public dashboard" wording is inaccurate | **Fixed** in `src/config.ts` Zod message + `src/start.ts` comment — now says "disable /status (returns 404)" |
| 3 | `.optional()` rejects YAML `null` (info) | **Fixed** via `.nullish().transform((v) => v ?? undefined)` on both `status.password` and `server.base_url`; added 6 tests covering `password:`, `password: null`, `password: ~`, quoted `"${VAR}"`, plus the parallel base_url cases |
| 4 | **blocking** — same `.optional()` vs null + missing curated error | **Fixed** (same as #3) |
| 5 | nit — `base_url:` (null) hits generic error | **Fixed** (same as #3) |
| 6 | **blocking/scope** — auto-gen widening vs #54 | **Fixed** (reverted; see above) |
| 7 | nit — `console.log` → `console.warn` | **Fixed**; warn now mentions both `BROKER_SIGNING_KEY_FILE` and `broker.signing_key_file` as override hints |
| 8 | nit — `mkdirSync(..., { mode: 0o700 })` for the parent dir | **Deferred** — the new code only `mkdir -p`s the legacy `<cwd>` parent (in practice the user's cwd or a tmpdir, already operator-controlled). Worth a follow-up once #54 lands and the auto-gen surface is gated. |
| 9 | nit — assert parent dir mode in test | **Deferred** (paired with #8) |
| 10 | nit — start.test coverage gap on real-start mkdir | **Not applicable post-revert** — the configured path no longer mkdirs; real-start cwd-fallback is covered by the new `tests/broker/signing.test.ts` case that exercises a synthetic cwd whose parent doesn't exist |

### DX nit (file-not-in-original-diff)
- `relay.yaml.example` now ships the `status:` block fully commented with an instructional block so `cp relay.yaml.example relay.yaml && dicode-relay` no longer hard-fails when `STATUS_PASSWORD` is unset. Uncommenting the two lines + setting the env var is the documented path to the enabled state.

### Follow-up
- A separate dicode-core PR (companion to #286) will handle first-run key creation in `buildin/relay-server` — i.e. core pre-creates the `${DICODE_DATADIR}/relay/` subdir and writes a P-256 PEM before spawning the relay child process. That removes the need for the relay to auto-gen at the YAML path at all, and aligns with the #54 narrowing direction.
- README update mentioning the auto-gen-at-YAML-path behavior (review nit 4 from the code review) is **not needed** post-revert — the configured-path branch no longer auto-generates, so the documented behavior matches the code again.

All 207 tests pass (`format:check`, `lint`, `build`, `test`).

dr14-make (commented):
## Round-2 review — APPROVE (comment-mode; GitHub blocks self-approval)

The force-pushed two-commit history (`a4468ad` + `2beb2ae`) cleanly addresses every round-1 finding. Diff is exactly what was described: scope trim is surgical, no collateral damage.

### Round-1 triage
| # | Finding | Status |
|---|---|---|
| 1 | mkdir widening / auto-gen at YAML path conflicts with #54 | **Fixed** — configured-path branch now throws ENOENT with a curated message naming the path; only the legacy cwd-fallback retains auto-gen, with the narrow `mkdir -p` safety the original brief intended. `tests/broker/signing.test.ts` asserts both behaviours, including that the throw path leaves *no* file or parent dir on disk. |
| 2 | `.optional()` rejects YAML `null` | **Fixed** — `status.password` and `server.base_url` both use `.nullish().transform((v) => v ?? undefined)`. 5 new tests pin each YAML null shape (`password:`, `password: null`, `password: ~`, quoted-empty, parallel `base_url`). |
| 3 | `relay.yaml.example` hard-fails fresh cp | **Fixed** — `status:` block fully commented with an instructional block explaining the dashboard's 404 default and the empty-value rejection. Fresh `cp && dicode-relay` no longer trips Zod. |
| 4 | README mention of new auto-gen behaviour | **N/A post-revert** — the configured-path branch no longer auto-generates, so the documented behaviour matches the code again. |
| 5 | `console.log` → `console.warn` | **Fixed** — `signing.ts:124` uses `console.warn` and now names both `BROKER_SIGNING_KEY_FILE` and `broker.signing_key_file` as override hints. |
| security-review info | "public dashboard" wording inaccurate (actually 404) | **Fixed** in both `src/config.ts` Zod message ("disable /status (returns 404)") and `src/start.ts` comment. |
| security nit | `mkdir mode 0o700` + parent-dir mode assertion | **Reasonably deferred** — the only remaining mkdir is on the legacy cwd-fallback (operator-controlled), and the broader auto-gen gating lives in #54. |
| coverage nit | real-start mkdir+write coverage | **Resolved post-revert** — `tests/broker/signing.test.ts` exercises a synthetic cwd whose parent doesn't exist, covering the real mkdir+write+0o600+stable-pubkey-on-second-load path. The startServer-integration test correctly asserts the throw path post-revert. |

### Scope-trim verification
- Two clean commits, well-scoped: `a4468ad` is just the `mkdir -p` safety on the legacy cwd path; `2beb2ae` is the Zod tightening + comment/error fixes + tests.
- No real coverage deleted: the previously-rewritten `start.test.ts` case has been *re-rewritten* back to the throw assertion (with a stricter regex match + a side-effects assertion that no file/dir was materialised). `tests/index.test.ts` got the same treatment for the CLI `--check` path.
- Legacy cwd-fallback issues (cwd writes, parallel-worker race per #71's round-2 review) are **not widened** by this PR — the only new behaviour on that branch is `mkdirSync(dirname(autoPath), { recursive: true })` immediately before the existing `writeFileSync`. Same write target, same race window.
- Diff stats (343+ / 31-): proportional to the Zod expansion + test additions. No drive-by churn.

### New issues introduced by the trim
None. The trim is conservative: it narrows the auto-gen surface (back to legacy cwd-only) and tightens the Zod schema. Both are net-positive for security and DX.

### Merge readiness
**Mergeable as-is.** The remaining hard dependency is the dicode-core supervisor hotfix (companion to #286) that pre-creates `${DICODE_DATADIR}/relay/` and writes a P-256 PEM before spawning the relay child — without it, the buildin/relay-server path will now ENOENT on first start instead of silently auto-generating. The author called this out in their reply; once that core PR lands (or is filed), this PR is ready to squash-merge. The two are complementary and the ordering (core hotfix → merge this) is correct.

Approving comment-mode.

dr14-make (commented):
## Round-2 security review

Verified the trim against round-1 findings. The auto-gen attack surface is **not** widened by this PR — the previously-flagged widening has been correctly reverted to a strict ENOENT throw, and the only auto-gen surface is the same legacy `<cwd>/broker-signing-key.pem` fallback that existed before this PR.

### Round-1 findings, status

| # | Finding | Status |
| --- | --- | --- |
| 1 | `mkdirSync` widens auto-gen write target to any YAML path | properly fixed — `signing_key_file` branch now throws ENOENT (`src/shared/signing.ts:87-93`); auto-gen confined to legacy cwd fallback (`:111-128`) |
| 2 | Comment/error claimed "absent → public dashboard" but `statusAuth(undefined)` returns 404 | properly fixed — `src/config.ts:111-117` and `src/start.ts:251-256` accurately describe 404 behavior |
| 3 | `.optional()` rejects YAML `password:` (null) with a generic Zod error | properly fixed — `.nullish().transform((v) => v ?? undefined)` applied to both `status.password` and `server.base_url`, with tests for `password:`, `password: null`, `password: ~` (`tests/config.test.ts:155-175`) |
| 4 | `console.log` for an operator-relevant side effect | properly fixed — switched to `console.warn` with updated message (`src/shared/signing.ts:124-127`) |
| 5 | `mkdirSync` parent dir uses umask, not `0o700` for a `0o600` key | acceptably deferred — author chose to defer; rationale (limited blast radius: only cwd fallback) is reasonable. Recommend filing a follow-up issue rather than blocking |
| 6 | Test only asserts `0o600` on file, not parent dir | acceptably deferred — paired with #5 |
| 7 | Loss of `startServer({ dryRun: false })` coverage | partially addressed — new `signing.test.ts` test covers the real-start auto-gen path with a synthetic cwd (verifies mkdir-p + `0o600` + persistence across calls); not via `startServer` itself, but functionally equivalent coverage |

### Verifications performed

- **`signing.ts:79-94` (configured-path branch):** confirms `existsSync` check throws before any disk write; no `mkdirSync` or `writeFileSync` reachable when `signingKeyFile !== ""` and the file is missing. ✓
- **`signing.ts:111-128` (legacy fallback):** `mkdirSync(dirname(autoPath), { recursive: true })` followed by `writeFileSync(autoPath, ..., { mode: 0o600 })`. Mode is `0o600` (private). ✓
- **dryRun guard (`start.ts:113`):** `allowAutoGenerate = opts.dryRun !== true` correctly suppresses disk writes in the legacy-fallback branch. Importantly, the ENOENT throw for configured paths fires *before* the `allowAutoGenerate` fork, so dryRun still surfaces the typo error correctly (tests confirm). ✓
- **`statusAuth(undefined)` semantics (`src/status/auth.ts:14-19`):** returns 404 handler; matches the new comments/errors. ✓
- **`relay.yaml.example`:** the `status:` block is fully commented out with explanatory comments that correctly state "omitting password makes /status return 404" and "empty value is rejected at load time". No dangerous direction. ✓
- **Test fixtures (`tests/`):** scanned for realistic-looking credentials — only placeholders found (`gh-secret`, `xoxb-resolved` mock prefix, `test-pw`-style). ✓
- **All `null` YAML forms (`password:`, `password: ~`, `password: null`):** explicit test coverage in `tests/config.test.ts:155-175`, all transform to `undefined`. ✓
- **Empty string still rejected (`.min(1)`):** explicit test at `tests/config.test.ts:144-146`. ✓
- **`${STATUS_PASSWORD}` unset → `""` → rejected:** explicit test at `tests/config.test.ts:120-141`. ✓

### New issues introduced by the rewrite

None. The Zod `.nullish().transform()` pattern is the idiomatic fix for the round-1 `null`-rejection bug, and the comments accurately document the resulting semantics. The mkdir-p change is now genuinely the narrow "first-start UX" fix the brief originally intended.

### Final verdict

**Approved from security standpoint.** The security posture is narrower for the YAML-configured path (typo → loud failure, no silent rotation) and unchanged for the legacy cwd fallback. The `status.password` and `server.base_url` Zod tightening is a strict improvement: empty/unresolved env interpolations now fail closed at load time with a curated message.

Recommend filing a low-priority follow-up issue tracking the `mkdir(mode: 0o700)` nit (round-1 finding #5) and #54's broader auto-gen hardening, but neither should block merge.

PR #75, opened 2026-05-10
refactor(relay): adopt lru-cache for NonceStore with 100k ceiling

dr14-make:
## Summary
- Replace hand-rolled `Map`+`setTimeout` in `NonceStore` with `LRUCache(max=100_000, ttl=ttlMs, ttlAutopurge=true)` to bound memory under a fresh-nonce flood.
- Mirrors the pattern landed for `SessionStore` in #56 — same `dateClock` shim so `vi.useFakeTimers()` keeps working (LRUCache's default clock captures `performance` at module load; fake timers replace the global wholesale).
- Public API (`check` / `size` / `clear`) unchanged; existing 60s-TTL and \`size\`-drops-to-zero assertions preserved by `ttlAutopurge: true`.
- Documents the replay-via-eviction edge: at sustained rates above `MAX_NONCES / ttlSeconds` (~1,667 handshakes/sec) the LRU would evict still-time-valid nonces, but the ±30s timestamp window in `RelayServer` provides the binding outer bound on replay validity.

## Context
This is one of three small refactors from the closed #58 (which had drifted too far from main to rebase). Companion PRs cover HTML escaping consolidation and the SPKI helper dedupe — filed separately for easier review.

## Test plan
- [x] `npm run typecheck`
- [x] `npm run lint`
- [x] `npm run format:check`
- [x] `npm run test` (208/208)
- [x] New eviction sanity test (`size` stays at or below ceiling under flood)

### DIFF
--- a/src/relay/nonces.ts
+++ b/src/relay/nonces.ts
@@ -1,19 +1,56 @@
 /**
- * NonceStore — prevents replay attacks by tracking seen nonces for 60 seconds.
- * O(1) lookup and insertion. Entries are evicted after their TTL expires.
+ * NonceStore — prevents replay attacks by tracking seen nonces for the
+ * configured TTL window (60 s in production). Backed by `lru-cache` so
+ * that an attacker flooding the handshake endpoint with fresh nonces
+ * cannot grow memory unbounded — entries are LRU-evicted at the ceiling.
+ *
+ * Replay-via-eviction: at sustained handshake rates above
+ * MAX_NONCES / ttlSeconds (≈ 1,667 handshakes/sec for 100k/60s) the LRU
+ * starts evicting non-expired nonces, which would in principle allow
+ * replay of an evicted but still-time-valid nonce. The handshake also
+ * enforces a ±30s timestamp window (see src/relay/server.ts), so a
+ * captured handshake whose nonce could be evicted is already too old to
+ * pass the timestamp check. The 60 s nonce TTL is the inner defense; the
+ * timestamp window is the binding outer bound on replay validity.
  */
 
-interface NonceEntry {
-  expiresAt: number;
-  timer: ReturnType<typeof setTimeout>;
-}
+import { LRUCache } from "lru-cache";
+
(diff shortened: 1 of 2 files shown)


dr14-make:
### Code review

Found 1 issue:

1. The new "flood" test does not actually exercise the LRU ceiling. It inserts 256 unique nonces and asserts `size <= 100_000` (trivially true) and `size === 256`, so eviction is never triggered and the documented invariant from the PR description ("`size` stays at or below ceiling under flood") is not verified. A future change that lowers or removes `max` would pass this test unchanged. The identical concern was raised on the predecessor PR ([#58 review comment on `nonces.ts`](https://github.com/dicode-ayo/dicode-relay/pull/58#discussion_r-flood)) with a concrete fix, and that PR's follow-up commit (`983cd72 review: address PR #58 feedback`) implemented it: insert `MAX_NONCES + 10` unique nonces with a long TTL (so eviction is purely LRU, not TTL), then assert `size === 100_000` and that the first-inserted nonce returns `false` on re-check (i.e. it was evicted). Suggest porting that test back in here.

https://github.com/dicode-ayo/dicode-relay/blob/047dd76c0a261c98a04ea752718e771ac0bc3362/tests/relay/handshake.test.ts#L356-L371

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

dr14-make:
### Follow-up review — fix verified

Commit `8d6d9fb` ports the stronger flood test verbatim from #58 commit `983cd72`:

- Inserts `MAX_NONCES + 10 = 100_010` unique nonces with a 1-hour TTL so eviction is purely LRU, not TTL.
- Asserts `size <= 100_000` (ceiling holds).
- Asserts `store.check(firstNonce)` returns `false` on a second call — proving the oldest entry was LRU-evicted rather than still present. Against an unbounded Map this would return `true` and fail loudly.

Locally `npm test` is green (27 files, 208/208). No other unaddressed nits from the prior pass — resolving.

PR #77, opened 2026-05-10
refactor(relay,shared): dedupe uncompressedP256ToSpki helper

dr14-make:
## Summary
- Both `src/shared/crypto.ts:81` and `src/relay/server.ts:517` carried byte-identical copies of `uncompressedP256ToSpki` (the 27-byte DER SPKI prefix for `ecPublicKey + prime256v1`).
- Export the canonical helper from `src/shared/crypto.ts` (where its sibling `verifyECDSA` already lives) and import it from `src/relay/server.ts`. Deletes the local copy and its section header.
- Crypto duplication is a footgun: a future tweak to the byte layout has to land in two places, and divergence produces silent verify failures with no compile-time signal. One canonical home eliminates that risk.
- The 27-byte SPKI header is part of the wire contract with the Go daemon; documented inline.

## Context
Third of three small refactors from the closed #58. Companions: #75 (NonceStore -> lru-cache), #76 (HTML escaping -> escape-html package).

When #58 was filed `src/shared/crypto.ts` didn't yet exist; the original PR proposed creating a new `src/shared/crypto-utils.ts`. Since then \`shared/\` was introduced organically and \`crypto.ts\` already owns one copy of the helper, so the cleanest landing is to export the existing copy rather than introduce a third file.

## Test plan
- [x] \`npm run typecheck\`
- [x] \`npm run lint\`
- [x] \`npm run format:check\`
- [x] \`npm run test\` (207/207)

### DIFF
--- a/src/relay/server.ts
+++ b/src/relay/server.ts
@@ -32,6 +32,7 @@ import {
   type Response as ResponseMessage,
   type ServerMessage,
 } from "./pb/relay_pb.js";
+import { uncompressedP256ToSpki } from "../shared/crypto.js";
 import type { ForwardResponse } from "../shared/protocol.js";
 
 // ---------------------------------------------------------------------------
@@ -507,17 +508,5 @@ export class RelayServer extends EventEmitter {
 }
 
 // ---------------------------------------------------------------------------
-// DER encoding helper
-// ---------------------------------------------------------------------------
-
-/**
- * Wraps a raw 65-byte uncompressed P-256 public key into a DER-encoded
- * SubjectPublicKeyInfo structure so Node.js crypto can import it.
- */
-function uncompressedP256ToSpki(pubkey: Buffer): Buffer {
-  const header = Buffer.from("3059301306072a8648ce3d020106082a8648ce3d030107034200", "hex");
-  return Buffer.concat([header, pubkey]);
-}
-
 // Re-export generated types for tests and external consumers.
 export type { Request, Response } from "./pb/relay_pb.js";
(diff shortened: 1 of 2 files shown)


dr14-make (commented):
Review (automated, confidence-filtered):

**Verdict:** LGTM, ready to merge.

**Byte-identity:** Confirmed — diffed both pre-PR function bodies at `origin/main`. The two copies produced identical `Buffer.concat([header, pubkey])` output; the only textual delta was a one-line inline comment (`// Fixed 27-byte SPKI header for ecPublicKey + prime256v1`) present in `shared/crypto.ts` but absent in `server.ts`. That comment is preserved (expanded into the JSDoc block) in the surviving export, and the new "wire contract with the Go daemon" note is a useful addition.

**Comment preservation:** OK. `server.ts`'s docstring ("Wraps a raw 65-byte uncompressed P-256 public key into a DER-encoded SubjectPublicKeyInfo structure so Node.js crypto can import it.") said the same thing as the surviving one, just with different prose. Nothing lost.

**Tests:** `tests/broker/crypto.test.ts` exercises `verifyECDSA` with valid sig / tampered payload / tampered sig / wrong pubkey / malformed pubkey (lines 118–163). Since `verifyECDSA` is the sole call site of `uncompressedP256ToSpki`, the helper is transitively covered. No new dedicated test needed for a pure 2-line constant-prefix wrapper.

**Typecheck/lint:** `npm run typecheck` and `npm run lint` both pass locally on `chore/spki-dedupe`.

**Nit (not blocking, confidence ~70 — below post threshold but flagging for the author):** The new `import { uncompressedP256ToSpki } from "../shared/crypto.js";` line lands at server.ts:35, separating the `./pb/relay_pb.js` block (line 25–34) from the existing `../shared/protocol.js` import (now line 36). Cleaner grouping would put both `../shared/*` imports adjacent. ESLint doesn't enforce `import/order` in this repo so it's purely cosmetic.

No security or correctness concerns. Helper is pure, deterministic, no side effects.

